PR #1, opened 2018-10-06
Proposal for JAva SDK

matzew:
Simple SDK for a Java API

### DIFF
--- a/api/src/main/java/io/cloudevents/CloudEvent.java
+++ b/api/src/main/java/io/cloudevents/CloudEvent.java
@@ -1,0 +1,77 @@
+/**
+ * Copyright 2018 The CloudEvents Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


matzew:
@ac360 I saw that you created this repo - So I figured it's a good time to get the Java SDK in.  

We talked about our (@hhiden, @sjwoodman and me) work in a few SDK calls

api/src/main/java/io/cloudevents/CloudEvent.java
```diff
@@ -0,0 +1,77 @@
...
+    String getEventID();
+
+    /**
+     * The version of the eventType. This enables the interpretation of data by eventual consumers, requires the consumer to be knowledgeable about the producer.
+     */
+    Optional<String> getEventTypeVersion();
+    /**
+     * Timestamp of when the event happened.
+     */
+    Optional<ZonedDateTime> getEventTime();
```
joschi: Shouldn't this be an `Instant` instead of `ZonedDateTime`?

> This class models a single instantaneous point on the time-line. This might be used to record event time-stamps in the application.

➡️ https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html

matzew: That was picked since the `Timestamp` (_String expression as defined in [RFC 3339](https://tools.ietf.org/html/rfc3339)_) can contain timezones.

I think that was suggested by @jroper on a mailing list thread a while ago

matzew:
@joschi I've merged this for now. to get the repo going.

Feel free to either open an issue for further discussions or a PR to discuss this as well. Thanks!

PR #9, opened 2018-11-12
Adding deserializer for ZonedDateTime

matzew:
support for Jackson's `StdDeserializer`, adding hooks for `ZonedDateTime` clazz

### DIFF
--- a/api/src/main/java/io/cloudevents/impl/ZonedDateTimeDeserializer.java
+++ b/api/src/main/java/io/cloudevents/impl/ZonedDateTimeDeserializer.java
@@ -1,0 +1,46 @@
+/**
+ * Copyright 2018 The CloudEvents Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


api/src/main/java/io/cloudevents/impl/ZonedDateTimeDeserializer.java
```diff
@@ -0,0 +1,46 @@
...
+        super(vc);
+    }
+
+    @Override
+    public ZonedDateTime deserialize(JsonParser jsonparser, DeserializationContext ctxt) throws IOException {
+        // not serializing timezone data yet
+        try {
+            return ZonedDateTime.parse(jsonparser.getText(), DateTimeFormatter.ISO_OFFSET_DATE_TIME);
+        } catch (DateTimeException e) {
+            throw new IllegalArgumentException("could not parse");
```
sjwoodman: Nit. Maybe include text that couldn't be parsed?

PR #12, opened 2018-11-21
Make the source compatible to java 1.8

fabiojose:
Signed-off-by: Fabio José <fabiojose@gmail.com>

In our company we have strong compliance rules and at this moment we are able to use just java 1.8.

Could you accept these changes to make the impl. compatible to java 1.8?

### DIFF
--- a/api/src/main/java/io/cloudevents/impl/ZonedDateTimeSerializer.java
+++ b/api/src/main/java/io/cloudevents/impl/ZonedDateTimeSerializer.java
@@ -1,0 +1,46 @@
+/**
+ * Copyright 2018 The CloudEvents Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.cloudevents.impl;
+
+import java.io.IOException;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+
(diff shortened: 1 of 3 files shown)


api/src/test/java/io/cloudevents/CloudEventBuilderTest.java
```diff
@@ -30,15 +31,18 @@
...
         final String eventId = UUID.randomUUID().toString();
         final URI src = URI.create("/trigger");
         final String eventType = "My.Cloud.Event.Type";
         final String eventTypeVersion = "2.0";
         final ZonedDateTime eventTime = ZonedDateTime.now();
         final String contentType = "application/json";
         final URI schemaUri = URI.create("http://cloudevents.io/schema");
-        final Map<String, String> extensionData = Map.of("foo", "bar");
+        final Map<String, String> extensionData = new HashMap<>();
+        extensionData.put("foo", "bar");
```
matzew: :crying_cat_face:   but agreed :smile:

api/src/main/java/io/cloudevents/impl/ZonedDateTimeSerializer.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.io.IOException;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+
+public class ZonedDateTimeSerializer extends StdSerializer<ZonedDateTime> {
```
matzew: thanks!

matzew:
@fabiojose thanks for the PR and the update!

api/src/test/java/io/cloudevents/CloudEventBuilderTest.java
fabiojose: The java9 solution to construct maps is amazing!

Any idea to release it?

api/src/main/java/io/cloudevents/impl/ZonedDateTimeSerializer.java
fabiojose: You are welcome!

PR #16, opened 2018-12-04
Vert.x-isms

jponge:


matzew:
Thanks for the PR and adding RxJava2 here @jponge

PR #17, opened 2018-12-04
Update the README against the new Vert.x interface

jponge:


matzew:
ah, thanks did oversee that.

PR #37, opened 2019-06-16
Ended request to write response back to caller

rinormaloku:
Without `req.response().end();` the event was hanging, i.e. if there were events every minute it was hanging on each for 5 minutes. 

It is something very simple but in my case, I copied trusting that the code is correct and then had to debug for hours to find out the problem.


fabiojose:
@rinormaloku Could you update your fork from upstream?

Issue #31, opened 2019-02-26
When adding a extension to a builder it is not being serialized to the message


maiconmattos:
https://github.com/cloudevents/sdk-java/blob/c85c691ea63516bfc7920ad4289ec1b6e267be4e/api/src/main/java/io/cloudevents/impl/DefaultCloudEventImpl.java#L36

When building a cloudEvent like below :
```
    final DistributedTracingExtension dte = new DistributedTracingExtension();
    dte.setTraceparent("00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01");
    dte.setTracestate("congo=BleGNlZWRzIHRohbCBwbGVhc3VyZS4");

    return new
        CloudEventBuilder<T>().contentType(ContentType.APPLICATION_JSON.getMimeType())
        .id("123")
        .time(ZonedDateTime.ofInstant(Instant.now(), ZoneOffset.UTC))
        .type("evType")
        .source("a.b.c")
        .data(entity)
        .extension(dte).build();
```

The extension is not being serialized when creating the json event.

maiconmattos:
Something like that could be implemented at DefaultCloudEventImpl to support Json serialization for extension values
```
private Map<String, Object> extendedAttributes = new HashMap<>();

  private void resolveExtensions(){
    for (Extension extension : this.getExtensions().orElse(new ArrayList<>())) {
      for (Field field : extension.getClass().getDeclaredFields()) {
        try {
          field.setAccessible(true);
          Object value = field.get(extension);
          extendedAttributes.put(field.getName(), value);
        } catch (IllegalAccessException e) {
          throw new RuntimeException(e);
        }
      }
    }
  }

  @JsonAnyGetter
  public Map<String, Object> getExtendedAttributes() {
    return extendedAttributes;
  }
```

tarunrathor-pro:
I also waiting for this issue to be addressed.

pajoma:
Any news on this? We need the extensions as well.

jmaister:
What is the reason to remove "eventTypeVersion" and "extensions" ? They are part of the event.

    @JsonIgnoreProperties(value = { "eventTypeVersion", "extensions" }) // was removed from 0.1

arichiardi:
We have also run against the problem and worked around a solution with a custom Jackson attribute, unfortunately this is not really ideal and we are looking into a better way.

razkevich:
We have found an easy workaround for this. So instead of doing this:

```
[...]

PR #43, opened 2019-08-14
To marshall and unmarshall extensions

fabiojose:
To fix the issue #31


pajoma:
Very good, we need this.

PR #65, opened 2019-10-24
Allow providing an external validator

ruromero:
Signed-off-by: Ruben Romero Montes <rromerom@redhat.com>

Allows providing an external Validator that would allow reusing an existing instance.
Hibernate Validator cannot be started properly in a native executable.

### DIFF
--- a/api/src/main/java/io/cloudevents/v03/CloudEventBuilder.java
+++ b/api/src/main/java/io/cloudevents/v03/CloudEventBuilder.java
@@ -42,8 +42,15 @@ import io.cloudevents.fun.EventBuilder;
  */
 public final class CloudEventBuilder<T> implements 
 		EventBuilder<T, AttributesImpl> {
-	private CloudEventBuilder() {}
-	
+
+	private CloudEventBuilder(Validator validator) {
+		if(validator == null) {
+			this.validator = getValidator();
+		} else {
+			this.validator = validator;
+		}
+	}
+
 	private static Validator VALIDATOR;
 	
 	public static final String SPEC_VERSION = "0.3";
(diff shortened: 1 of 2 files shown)


ruromero:
@fabiojose do you mind reviewing this PR?
Thanks

fabiojose:
@ruromero Please, provide some unit tests.

ruromero:
@fabiojose done, thanks

gunnarmorling:
> Hibernate Validator cannot be started properly in a native executable.

@ruromero, can you expand on this one? What is the problem specifically? HV works in GraalVM native images via Quarkus in general, but indeed there are some limitations atm. (e.g. no XML support). So I'm curious which issue you ran into. Thanks!

ruromero:
In Quarkus the static initialisation of the Validator was causing an error at runtime. This way we can use the instance provided by CDI.
See also the follow up PR #73

SchulteMarkus:
> In Quarkus the static initialisation of the Validator was causing an error at runtime. This way we can use the instance provided by CDI.
> See also the follow up PR #73

Because of this change I was able to use cloudevents-api:io.cloudevents as part of my Quarkus native application, nice one!

```java
@Inject Validator validator;
...
final CloudEvent<AttributesImpl, Map> cloudEvent 
  = io.cloudevents.v03.http.Unmarshallers.binary(Map.class, validator)
```

But well, as a user of cloudevents-api:io.cloudevents it would be even more comfortable to not care this issue.

SchulteMarkus:
@ruromero Maybe we should fix the reason for this issue (for using native-image, having to inject of validator-instance manually). But I don't know where to start. Is the core issue located in cloudevents-api:io.cloudevents, or should Quarkus be patched?

ruromero:
@SchulteMarkus The only way to make this more "quarkus-user-friendly" is to create a Quarkus extension but I will need to ask the Quarkus team to help me out on this one :smile:
I can create the issue on Quarkus and link it here so more people can up-vote it.
What do you think?

SchulteMarkus:
@ruromero Please create an corresponding issue at the Quarkus board.

Issue #70, opened 2019-11-12
Implement "dataref" extension


paulschwarz:
Do you plan to implement the dataref extension?

https://github.com/cloudevents/spec/blob/master/extensions/dataref.md

PR #71, opened 2019-11-12
#70 Implement dataref extension

paulschwarz:
See issue #70 

I have implemented the `dataref` extension based on the spec https://github.com/cloudevents/spec/blob/master/extensions/dataref.md and following the implementation for the tracing extension already provided in this SDK.

### DIFF
--- a/api/src/main/java/io/cloudevents/extensions/DatarefExtension.java
+++ b/api/src/main/java/io/cloudevents/extensions/DatarefExtension.java
@@ -1,0 +1,110 @@
+package io.cloudevents.extensions;
+
+import java.net.URI;
+import java.util.AbstractMap.SimpleEntry;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+public class DatarefExtension {
+
+  private URI dataref;
+
+  public URI getDataref() {
+    return dataref;
+  }
+
+  public void setDataref(URI dataref) {
+    this.dataref = dataref;
+  }
+
+  @Override
+  public String toString() {
+    return "DatarefExtension{" +
+        "dataref='" + dataref + '\'' +
+        '}';
+  }
+
+  @Override
+  public int hashCode() {
+    final int prime = 31;
+    int result = 1;
+    result = prime * result + ((dataref == null) ? 0
+        : dataref.hashCode());
+    return result;
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) {
+      return true;
+    }
+    if (obj == null) {
+      return false;
+    }
+    if (getClass() != obj.getClass()) {
+      return false;
+    }
+    DatarefExtension other = (DatarefExtension) obj;
+    if (dataref == null) {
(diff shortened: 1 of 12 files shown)


paulschwarz:
**NEED HELP**

Actually, this pull request is wrong. Notice how `dataref` is double nested? (similar to `distributedTracing`. I believe both of these are wrong. I see nothing in the spec that indicates that this should be the case.

Anyway, back to `dataref`...

```
{
    ...
    "dataref": {
        "dataref": "http://localhost/foo/bar/123"
    },
    "distributedTracing": {
        "traceparent": "foo",
        "tracestate": null
    }
}
```

According to the spec https://github.com/cloudevents/spec/blob/master/extensions/dataref.md it should look like this

```
{
    ...
    "dataref": "http://localhost/foo/bar/123"
    ...
}
```

However, I can't see how the library is building the json and I don't see a way to solve this. Please help, or let me know how. @fabiojose

paulschwarz:
I've looked at other SDK implementations, and my argument seems to have been validated. Extensions should not be nested as seen above in `distributedTracing`.

C# example: https://github.com/cloudevents/sdk-csharp/blob/527ab33deba6a0b3eb0d17640b02e6b0012a49dd/test/CloudNative.CloudEvents.UnitTests/ExtensionsTest.cs#L22-L23

JS example: https://github.com/cloudevents/sdk-javascript/blob/master/examples/payload/v1/structured-event-2.json#L9

fabiojose:
Let's check!

paulschwarz:
Did you get a chance to look at this?

What are the facts you're considering (so I can try to help you)?

fabiojose:
@paulschwarz 
You can do something like that: 

This is just an example, not suitable for unmarshalling.

```java
    public class DatarefExtension implements ExtensionFormat {
		
		public static final String KEY = "dataref";

		private final Map<String, String> transport = new HashMap<>();
		private final InMemoryFormat inmemory;
		
		public DatarefExtension(final String value) {
			transport.put(KEY, value);
			
			this.inmemory = new InMemoryFormat() {
				
				@Override
				public Class<?> getValueType() {
					return String.class;
				}
				
				@Override
				public Object getValue() {
					return value;
				}
				
				@Override
				public String getKey() {
					return KEY;
				}
			};
		}
		
		@Override
		public InMemoryFormat memory() {
			return inmemory;
		}

		@Override
		public Map<String, String> transport() {
			return transport;
		}
	}
```

slinkydeveloper:
Hi @paulschwarz, we are working on v2 of this sdk and we made some drastic changes, eventually breaking your PR :smile: Now implementation of an extension should be far more simple (look at the distributed tracing extension implementation): you just need to implement an interface and that's all. Are you willing to rebase this pr and reimplement this extension?

paulschwarz:
@slinkydeveloper thanks for the info. Given the issues I described above in this thread, I was uncomfortable with my PR, so I'm glad to hear that v2 promises to make extension simpler. I will take a look soon, time permitting. Thanks for the bump.

paulschwarz:
See #151

PR #79, opened 2019-12-03
Fix base64 marshalling

ruromero:
The base64 as it is described in the spec should be contained in a data_base64 attribute regardless of the type that's why I created a new param.

I have also remove it from the extensions and now it should be properly marshalled/unmarshalled (see tests).

Let me know if this PR is aligned with what you expect this field to work.

I also added some more clean up. Can't help it :smile:

### DIFF
--- a/api/src/main/java/io/cloudevents/CloudEvent.java
+++ b/api/src/main/java/io/cloudevents/CloudEvent.java
@@ -37,7 +37,9 @@ public interface CloudEvent<A extends Attributes, T> {
 	 * The event data
 	 */
 	Optional<T> getData();
-	
+
+	byte[] getDataBase64();
+
 	/**
 	 * The event extensions
 	 */
--- a/api/src/main/java/io/cloudevents/CloudEvent.java
+++ b/api/src/main/java/io/cloudevents/CloudEvent.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2019 The CloudEvents Authors
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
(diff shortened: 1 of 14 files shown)


ruromero:
@fabiojose note the change in the violations Set. The validator returns an emtpySet and this set is immutable.

api/src/main/java/io/cloudevents/CloudEvent.java
```diff
@@ -20,27 +20,26 @@
...
+     * The event context attributes
+     */
+    A getAttributes();
+
+    /**
+     * The event data
+     */
+    Optional<T> getData();
+
+    byte[] getDataBase64();
```
fabiojose: Why we can not use `Optional<T> getData()`?

Why an attribute called getDataBase64 exclusively to work with base64?

fabiojose:
In general, for me, keep using getData(), withData(), will be provided a more fluid use of the SDK.

api/src/main/java/io/cloudevents/CloudEvent.java
ruromero: Because there's no way to assign a `byte[]` to `data` as it is using generics. In the tests you can see that instead of declaring `Optional<byte[]>` I declare `Optional<String>`. Theoretically the `byte[]` should be a binary representation of the String and maybe the SDK consumer might want to create the `String` from the `byte[]`.

When receiving a structured CloudEvent with content type being `text/plain`, it can include a `data_base64` that I won't be able to easily unmarshal into a `CloudEvent<? extends Attributes, String>` that should also work for a CloudEvent with `data` being just text. Do you see my point? Am I missing something?

Thanks for your comments.

fabiojose:
@ruromero, provide the changelog.

Editing the CHANGELOG.md

PR #80, opened 2019-12-03
Replace ZonedDateTime with valid String representation

### LINKED ISSUE #29: ZonedDateTime field limits interoperability
Currently the `time` field is a `ZonedDateTime`. This requires a custom deserializer `ZonedDateTimeDeserializer` implemented with jackson.  Sending a CloudEvent via http then causes problems on the receiving end.  IMHO It would be better to represent time with a more portable type, e.g., long.

dturanski:
Resolves #29


dturanski:
Hold up on this. I'm working on some tweaks.

dturanski:
I pushed another commit to make it better

fabiojose:
@dturanski 

Please, sign off the commits.

api/src/test/java/io/cloudevents/v1/CloudEventJacksonTest.java
```diff
@@ -345,5 +341,4 @@ public void should_marshal_data_byte_array_as_data_base64() {
 		assertTrue(encoded.contains("\"data_base64\""));
 		assertTrue(encoded.contains("\"" + expected +"\""));
 	}
-
```
fabiojose: Please, avoid pushing files just with empty line remove.

api/src/test/java/io/cloudevents/v02/CloudEventJacksonTest.java
```diff
@@ -20,9 +20,15 @@
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+import com.google.gson.Gson;
```
fabiojose: I did not able to see where you are using this import.

api/src/main/java/io/cloudevents/v03/AttributesImpl.java
```diff
@@ -30,19 +35,15 @@
 import javax.validation.constraints.Pattern;
 import javax.validation.constraints.Size;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonInclude.Include;
-import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
```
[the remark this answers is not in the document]
dturanski: Sorry about that. Do you have checkstyle rules or anything?

dturanski:
The current test suite verifies that we didn't break anything.  Note that the `gson` import is left over from an attempt to test improved interoperability. Interoperability motivated this PR, but as I commented in the issue, this fix doesn't help with that since there are additional Jackson dependencies. My use case involved previous versions of cloud events and and the `projectriff/java-function-invoker` which was using gson internally to deserialize json payloads. As it turns out, the JFI uses Jackson now. GSon still doesn't work for other reasons. All attributes are null for example. So maybe the additional test is that I should be able to use `ObjectMapper` out of the box with no special config vs.  requiring `JSon`.  I'll add that.

slinkydeveloper:
Thank you for your contribution, we are reworking the sdk and this PR now is invalid. Look here: https://github.com/cloudevents/sdk-java/issues/29#issuecomment-620441035

PR #107, opened 2020-03-20
Vertx generic data

### LINKED ISSUE #106: http-vertx readFromRequest and writeToHttpClientRequest only allow CE with dataType=String
I think it should be generic.

matejvasek:
fixes #106


fabiojose:
@matejvasek please, change this PR just for generics types and use another one to discuss the version update.

matejvasek:
@fabiojose done

PR #108, opened 2020-03-23
Rework the Event interface

slinkydeveloper:
Hi,
I wish to propose a rework of the `Event` interface and related. I've tried to use this sdk in a middleware experiment (https://github.com/slinkydeveloper/cloud-streaming-data-plane) and I've found several issues with the actual `Event` interface.

I think the problems from a middleware point of view are:

* I don't care about the attributes version and I don't know it in advance, I just want to check if the event is valid and route it.
* I don't need to map the payload of the event, it's an expensive operation for me.

While from an "end user" perspective (who reads and writes events from json or from http):

* I usually don't care about the attributes version and I don't know it in advance, I just want to access to the basic metadata (id, type, time, etc). When I map back and forth to json I usually have no clue what is the attributes version
* Because of the "polyglot" aspect of the `data` field, I don't want to handle the difference between base64 and json serialized. I just want to get the `data` in the way I want (as json or as binary array)

From the experience of working on sdk-go and sdk-rust I've tried to bake a modified `Event` interface, let me know your thoughts.


matzew:
/lgtm

cc @fabiojose @jponge

slinkydeveloper:
Although the languages are really different, I want to point how the structure is designed in sdk-rust: 

* [Event](https://github.com/cloudevents/sdk-rust/blob/master/src/event/event.rs)
* [Data union type](https://github.com/cloudevents/sdk-rust/blob/master/src/event/data.rs)
* [Attributes union type](https://github.com/cloudevents/sdk-rust/blob/master/src/event/attributes.rs)
* [Extension value union type](https://github.com/cloudevents/sdk-rust/blob/master/src/event/extensions.rs)

fabiojose:
@slinkydeveloper remember to signoff the commits

fabiojose:
I would like to propose to remove any external dep, like Jackson and Bean Validations.

With that, CloudEvents API will work with any framework without issues.

fabiojose:
> Although the languages are really different, I want to point how the structure is designed in sdk-rust:
> 
>     * [Event](https://github.com/cloudevents/sdk-rust/blob/master/src/event/event.rs)
> 
>     * [Data union type](https://github.com/cloudevents/sdk-rust/blob/master/src/event/data.rs)
> 
>     * [Attributes union type](https://github.com/cloudevents/sdk-rust/blob/master/src/event/attributes.rs)
> 
>     * [Extension value union type](https://github.com/cloudevents/sdk-rust/blob/master/src/event/extensions.rs)

@slinkydeveloper  is really hard to create the same structure in every SDK.

fabiojose:
Go ahead! it's looking nice!

slinkydeveloper:
Ok I'll continue down that road, but be aware that this means reworking most of the sdk. Plus I'm going to drop 0.2, because of its weird extension things (and because according to this document we don't need it https://github.com/cloudevents/spec/blob/master/SDK.md#contribution-acceptance

fabiojose:
@slinkydeveloper consider these issues #111, #112 , #113 , #114 , #115 . This PR would be the start of really good modularization.

slinkydeveloper:
@fabiojose since those changes are radical, i think we need a v2 version of the sdk for this. Do you agree on branching the current master creating a 1.x branch?

api/src/main/java/io/cloudevents/CloudEvent.java
```diff
@@ -15,31 +15,52 @@
...
 
 /**
  * An abstract event envelope
- * @param <A> The attributes type
- * @param <T> The 'data' type
  * @author fabiojose
+ * @author slinkydeveloper
  */
-public interface CloudEvent<A extends Attributes, T> {
+public interface CloudEvent {
```
n3wscott, line 28: yes this is a good change. too many generics makes this really hard to use.

api/src/main/java/io/cloudevents/format/BinaryUnmarshaller.java
```diff
@@ -133,9 +133,9 @@ private BinaryUnmarshaller() {}
 		 */
 		HeadersStep<A, T, P> builder(EventBuilder<T, A> builder);
 	}
-	
+
 	private static final class Builder<A extends Attributes, T, P> implements
```
n3wscott, line 137: this still has the layer cake of generics

api/src/main/java/io/cloudevents/json/CloudEventSerializer.java
```diff
@@ -0,0 +1,24 @@
...
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import io.cloudevents.CloudEvent;
+
+import java.io.IOException;
+
+public class CloudEventSerializer extends StdSerializer<CloudEvent> {
```
n3wscott, line 11: nice. nice nice.

api/src/main/java/io/cloudevents/json/Json.java
```diff
@@ -15,35 +15,53 @@
...
-
-        final SimpleModule module = new SimpleModule();
+        // add ZonedDateTime ser/de
+        final SimpleModule module = new SimpleModule("Custom ZonedDateTime");
         module.addSerializer(ZonedDateTime.class, new ZonedDateTimeSerializer());
         module.addDeserializer(ZonedDateTime.class, new ZonedDateTimeDeserializer());
         MAPPER.registerModule(module);
     }
 
+    protected static final DateTimeFormatter RFC3339_DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ssXXX");
```
n3wscott, line 45: for what it's worth, the go sdk uses the RFC3339 Nano format... but they both work together.

api/src/main/java/io/cloudevents/json/Json.java
```diff
@@ -15,35 +15,53 @@
...
+
+    public static String encode(final CloudEvent event) throws IllegalStateException {
```
n3wscott, line 47: encode == structured mode?

n3wscott:
Here is my little demo of using the current java sdk in a spring app and it was quite painful:

https://github.com/n3wscott/spring-ce-demo/blob/master/src/main/java/com/n3wscott/demo/DemoApplication.java

I would love to see the amount of work I need to do be reduced. This PR is starting in that direction, Thanks @slinkydeveloper !!

api/src/main/java/io/cloudevents/format/BinaryUnmarshaller.java
slinkydeveloper: This still needs to be removed :smile:

api/src/main/java/io/cloudevents/json/Json.java
slinkydeveloper: So these are marshal/unmarshal methods for Cloudevent to go back and forth to JsonValue/String encoded json/binary encoded json

api/src/main/java/io/cloudevents/CloudEvent.java
```diff
@@ -15,31 +15,62 @@
...
+
+    /**
+     * Write an extension into this cloud event
+     * @param e
+     */
+    default void writeExtension(Extension e) {
+        e.writeToEvent(this);
+    }
+
+    static io.cloudevents.v1.CloudEventBuilder build() {
```
bsideup, line 65: this method should either return a common, non-versioned `CloudEventBuilder` interface, or be removed.

Otherwise, if for example spec `1.1` is released, changing this method to `io.cloudevents.v1_1.CloudEventBuilder build` would be a binary breaking change, and I expect it to be the most used way of creating the event.

api/src/main/java/io/cloudevents/CloudEvent.java
```diff
@@ -15,31 +15,62 @@
...
+     *
+     * Extensions values could be String/Number/Boolean
      */
     Map<String, Object> getExtensions();
+
+    /**
+     * Write an extension into this cloud event
+     * @param e
+     */
+    default void writeExtension(Extension e) {
```
bsideup, line 61: this brings mutability to this important interface.

It would be better to have it as `CloudEvent withExtension(Extension e)` that returns a new one

slinkydeveloper: Now that i see it, i prefer to have it removed since i designed the `CloudEvent` to be immutable

api/src/main/java/io/cloudevents/CloudEvent.java
```diff
@@ -15,31 +15,62 @@
...
-    byte[] getDataBase64();
+    /**
+     * The event data
+     */
+    Optional<byte[]> getDataAsBytes() throws DataConversionException;
+
+    /**
+     * The event data
+     */
+    Optional<JsonNode> getDataAsJson() throws DataConversionException;
```
bsideup, line 48: since this is a shortcut to `event.getDataAsString().map(objectMapper::readTree)`, I'd advice against adding a strong dependency on Jackson to the public API.

It is fine to use Jackson internally, but, if we add it to the public API, we will stick to it for a looooooooong time

api/src/main/java/io/cloudevents/CloudEvent.java
bsideup: `event.toBuilder().extension(e).build()` is an option too

api/src/main/java/io/cloudevents/SpecVersion.java
```diff
@@ -0,0 +1,25 @@
...
+    @Override
+    public String toString() {
+        return this.stringValue;
+    }
+
+    public static SpecVersion parse(String sv) {
+        switch (sv) {
+            case "0.3": return SpecVersion.V03;
+            case "1.0": return SpecVersion.V1;
+            default: throw new IllegalArgumentException("Unrecognized SpecVersion "+ sv);
```
bsideup, line 22:
```suggestion
            default: throw new IllegalArgumentException("Unrecognized SpecVersion " + sv);
```

api/src/main/java/io/cloudevents/extensions/DistributedTracingExtension.java
```diff
@@ -1,16 +1,18 @@
...
-public class DistributedTracingExtension {
+public final class DistributedTracingExtension implements Extension {
+
+    public static final String TRACEPARENT = "traceparent";
+    public static final String TRACESTATE = "tracestate";
 
     private String traceparent;
     private String tracestate;
 
+    public DistributedTracingExtension() { }
```
bsideup, line 14: Java classes already have it, implicitly

api/src/main/java/io/cloudevents/extensions/DistributedTracingExtension.java
```diff
@@ -27,21 +29,43 @@ public void setTracestate(String tracestate) {
...
+        }
+    }
+
+    @Override
+    public void writeToEvent(CloudEvent event) {
+        if (traceparent != null) {
+            event.getExtensions().put(TRACEPARENT, this.traceparent);
+        }
+        if (tracestate != null) {
+            event.getExtensions().put(TRACESTATE, this.tracestate);
```
bsideup, line 50: `event.getExtensions()` should really return a non-modifiable `Map` here

api/src/main/java/io/cloudevents/CloudEvent.java
slinkydeveloper: It's not a shortuct for `getDataAsString()`, the reason to have it is that when the event is created from the json representation, there are events where the payload is already a json (like the third example here https://github.com/cloudevents/spec/blob/v1.0/json-format.md#32-examples). This allows to skip the marshalling to string and back. 

Be aware that in any case we'll have a dependency on jackson.

api/src/main/java/io/cloudevents/extensions/ExtensionsParser.java
```diff
@@ -0,0 +1,41 @@
...
+import io.cloudevents.CloudEvent;
+import io.cloudevents.Extension;
+
+import java.util.HashMap;
+import java.util.function.Supplier;
+
+public final class ExtensionsParser {
+
+    private static class SingletonContainer {
+        private final static ExtensionsParser INSTANCE = new ExtensionsParser();
```
bsideup, line 12: `ExtensionsParser` as `enum` with a single `INSTANCE` value would work as well

slinkydeveloper: TBH I prefer this way, the singletons with enums sounds too weird to me :smile:

api/src/main/java/io/cloudevents/CloudEvent.java
```diff
@@ -15,31 +15,62 @@
...
+ * @author slinkydeveloper
  */
-public interface CloudEvent<A extends Attributes, T> {
+public interface CloudEvent {
 
     /**
      * The event context attributes
      */
-    A getAttributes();
+    Attributes getAttributes();
```
bsideup, line 33: FYI https://github.com/cloudevents/sdk-java/issues/118

api/src/main/java/io/cloudevents/Attributes.java
```diff
@@ -16,50 +16,48 @@
...
+     * TODO
 	 * A common way to get the media type of CloudEvents 'data';
 	 * @return If has a value, it MUST follows the <a href="https://tools.ietf.org/html/rfc2046">RFC2046</a>
 	 */
-	@JsonIgnore
-	Optional<String> getMediaType();
-	
+	Optional<String> getDataContentType();
+
+	Optional<URI> getDataSchema();
```
bsideup, line 57: Just my 2c:
`Optional` is a nice API, but it is costly (allocation wise). I wish we could sacrifice a bit of UX here in favour of performance and use `@Nullable`.

Having it as `Optional` would make it impossible to avoid allocations in third party performance critical implementations

api/src/main/java/io/cloudevents/CloudEvent.java
bsideup: cc @n3wscott

api/src/main/java/io/cloudevents/Attributes.java
slinkydeveloper: I'm trying to build 2 level APIs here like i did in sdk-go and sdk-rust:

* Event API: safer, more costly
* Message API: optimized for transferring events

api/src/main/java/io/cloudevents/CloudEvent.java
bsideup: dependency - yes, but Jackson type in the main public API...

Have you considered doing something like `<T> Optional<T> getDataAs(Class<T> clazz)`?

```java
event.getDataAs(JsonNode.class);
event.getDataAs(String.class);
event.getDataAs(byte[].class);
event.getDataAs(ByteBuffer.class);
```

would be highly optimizable on the implementation side while keeping the API JDK-only.

api/src/main/java/io/cloudevents/CloudEvent.java
slinkydeveloper: I think the solutions here are two:

* CloudEvent is implemented once (`impl.CloudEventImp`) and different attributes versions are added through composition of objects.
* CloudEvent is implemented multiple times, one for each event version.

The reason why i went with the first alternative is to simplify the implementation of json marshal/unmarshal, let's review this comment later when this pr is ready.

api/src/main/java/io/cloudevents/DataConversionException.java
```diff
@@ -0,0 +1,8 @@
+package io.cloudevents;
+
+public class DataConversionException extends RuntimeException {
+
+    public DataConversionException(String from, String to, Throwable cause) {
```
bsideup, line 5:
```suggestion
    public DataConversionException(Object from, Object to, Throwable cause) {
```

would allow `throw new DataConversionException(somethingNullable, someOtherNullable, ex)`

api/src/main/java/io/cloudevents/Attributes.java
slinkydeveloper: where `@Nullable` annotation comes from?

api/src/main/java/io/cloudevents/extensions/ExtensionsParser.java
bsideup: Actually... is it possible to avoid singletons for extensions by passing an extension parser to the event parser

It would be nice to be able to provide non-static extension parsers (e.g. by injecting with some DI framework)

api/src/main/java/io/cloudevents/extensions/ExtensionsParser.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public static ExtensionsParser getInstance() {
+        return SingletonContainer.INSTANCE;
+    }
+
+    private HashMap<Class<?>, Supplier<Extension>> extensionFactories;
```
bsideup, line 19: class -> impl is known to be hard to work with.

Have you considered a set of factories with `supports(Class<T> extensionClass)` method?

api/src/main/java/io/cloudevents/extensions/ExtensionsParser.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @SuppressWarnings("unchecked")
+    public <T extends Extension> T parseExtension(Class<T> extensionClass, CloudEvent event) {
+        Supplier<Extension> factory = extensionFactories.get(extensionClass);
+        if (factory != null) {
+            Extension ext = factory.get();
+            ext.readFromEvent(event);
+            return (T) ext;
+        }
+        return null;
```
bsideup, line 38: this `null` is ambiguous ("factory returned null" or "unknown factory")

api/src/main/java/io/cloudevents/format/EventFormatProvider.java
```diff
@@ -0,0 +1,34 @@
...
+
+    private static class SingletonContainer {
+        private final static EventFormatProvider INSTANCE = new EventFormatProvider();
+    }
+
+    public static EventFormatProvider getInstance() {
+        return EventFormatProvider.SingletonContainer.INSTANCE;
+    }
+
+    private HashMap<String, EventFormat> formats;
```
bsideup, line 17: `String` as a key is hard to work with, given that it can be something like "application/json; charset=utf-8" (the current version of SDK's HTTP support fails because of that "; charset=utf-8", btw)

api/src/main/java/io/cloudevents/Attributes.java
bsideup: it should come from `io.cloudevents` or a sub-package and be a JSR-305 meta-annotation, similar to:
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/lang/Nullable.html

bsideup: > I'm trying to build 2 level APIs here

I just submitted an issue to discuss before we implement it:
https://github.com/cloudevents/sdk-java/issues/119

api/src/main/java/io/cloudevents/common/BinaryData.java
```diff
@@ -0,0 +1,58 @@
+package io.cloudevents.common;
+
+import com.fasterxml.jackson.databind.JsonNode;
```
fabiojose: It would be nice to remove all external deps from API. This will avoid any kind of conflict with pre-existing libs or frameworks.

When someone wants to use JSON, they are concerned about Jackson usage. Checkout this issue #113.

api/src/main/java/io/cloudevents/format/json/CloudEventSerializer.java
```diff
@@ -0,0 +1,23 @@
+package io.cloudevents.format.json;
+
+import com.fasterxml.jackson.core.JsonGenerator;
```
fabiojose, line 3: What do you think about move Jackson dependencies to json-jackson module? Check out this issue #113

fabiojose: Without any external dependency in the API module, anyone who wants to use the API is free of conflicts with pre-existing frameworks and libraries.

bsideup: also see https://github.com/cloudevents/sdk-java/pull/108#discussion_r411153130

api/src/main/java/io/cloudevents/common/BinaryData.java
slinkydeveloper: look this thread https://github.com/cloudevents/sdk-java/pull/108#discussion_r411153130

api/src/main/java/io/cloudevents/CloudEvent.java
slinkydeveloper: Yeah that's what i did :)

api/src/main/java/io/cloudevents/CloudEvent.java
slinkydeveloper: To be honest i wish to avoid having a method like this because it could imply to have some encoding/decoding mechanism for the payload, similar to what we have in sdk-go (datacodec package).
TBH I'm prone to just have the payload as `[]byte`. It's easier both for the sdk and the users. It removes the optimization with json, but i think it's fine.

slinkydeveloper:
@fabiojose i've create a branch for sdk 1.x https://github.com/cloudevents/sdk-java/tree/1.x

api/src/main/java/io/cloudevents/CloudEvent.java
slinkydeveloper: yes now that i look at it i think it'a just better to have `byte[]` as data type

api/src/main/java/io/cloudevents/SpecVersion.java
slinkydeveloper: Mh i'm wondering why the intellij refactoring didn't spotted this one...

api/src/main/java/io/cloudevents/extensions/ExtensionsParser.java
slinkydeveloper: I think as a first pass is fine, we'll improve this part later. For now i just made the bare minimum to make it working, since it's my priority :smile:.

api/src/main/java/io/cloudevents/extensions/ExtensionsParser.java
slinkydeveloper: Why an extension factory should return null?

api/src/main/java/io/cloudevents/format/EventFormatProvider.java
slinkydeveloper: I think a correct usage should invoke `resolveFormat` with a parsed content type header (so without passing `charset`)

pom.xml
```diff
@@ -68,6 +68,7 @@
 
     <modules>
         <module>api</module>
+        <module>formats/json-jackson</module>
```
fabiojose, line 71: Another thing nice to see! #113

slinkydeveloper: Yep, now `cloudevents-api` depends on nothing!

olegz:
I have a more general question. Why do we need to make `*-api` module bringing so many dependency that are not really relevant to the task in hand. What is the task in hand? IMHO - _to define strategies to support the specification_. We should be free to write parsers, readers, writers, validators etc., if we choose to without bringing unnecessary dependencies which may actually create  conflict. Also see - https://github.com/cloudevents/sdk-java/issues/111

At it's current state `cloudevents-api-1.0.3` already brings the following dependencies

```
jackson-core-2.10.1.jar
jackson-databind-2.10.1.jar
jackson-annotations-2.10.1.jar
jackson-datatype-jdk8-2.10.1.jar
hibernate-validator-6.0.17.Final.jar
validation-api-2.0.1.Final.jar
jboss-logging-3.3.2.Final.jar
classmate-1.3.4.jar
jakarta.el-3.0.3.jar
```
Why. . .? 
There is a specification which essentially defines a contract and that should be enough for anyone to consume/produce CloudEvent types and that is all that such module should contain. 
Sure, as a community we can collaborate on implementation for _dealing_ with these events (e.g., marshalers, unmarshalers, etc. . ), but that should be a completely separate effort... may be even several efforts given the infancy of the current API.

We certainly don't wan to get into arguments such as _". . .this marshaller is more performant than that,. . . Gson vs Jackson etc. . .,  "_

bsideup:
@olegz FYI this PR seems to remove the dependencies from -api:
https://github.com/cloudevents/sdk-java/pull/108#discussion_r411501439

slinkydeveloper:
> There wasn't enough (public?) discussions on issues like this one:

This discussion started the 27 Feb 2020 in sdk meetings (https://docs.google.com/document/d/1JIZxnV_w-dMrSAguG4bqog1vBlD94mm5aFjJlfb-R34/edit?usp=sharing) but nobody had counter-arguments again a big refactor of the sdk through a pr like this.

As I said, I don't like proceeding this way and I'm ready to engage further discussions (like `Optional`s and `@Nullable`s). This is just a first pass, to clean up some stuff we don't need any more because of 0.2 and, most important, to align the APIs to other SDKs, bringing the `Event`/`Message` levels to provide on one side easy to use apis to read/write events and, on the other side, a simple api to marshal/unmarshal events back and forth the wire.

It's not perfect: on the contrary **it needs a lot of work and there are no definitive decisions**, but we need to start somewhere. Follow-up PRs will improve the work done here and, i underline, i'm really happy to engage the community to discuss these improvements.

slinkydeveloper:
I'm going to also open a Github project with things we want to have, to discuss and to improve for sdk v2

bsideup:
@slinkydeveloper 
>  This discussion started the 27 Feb 2020 in sdk meetings (https://docs.google.com/document/d/1JIZxnV_w-dMrSAguG4bqog1vBlD94mm5aFjJlfb-R34/edit?usp=sharing) but nobody had counter-arguments again a big refactor of the sdk through a pr like this.

I am sure there were _some_ discussions, but it would be better if they were more asynchronous (e.g. on GitHub issues) because not everyone can join the call and have the answers ready :)

> It's not perfect: on the contrary it needs a lot of work and there are no definitive decisions, but we need to start somewhere.

I hope my comments did not sound attacking - that's a great change and I am glad to see it happening! Just wanted to mention (as a feedback) that, as an external collaborator, I find the _way_ the SDK gets such changes a bit closed group-ed.

> I'm going to also open a Github project with things we want to have, to discuss and to improve for sdk v2

That's a great idea! It will definitely help the visibility of both the current and upcoming states 👍

api/src/main/java/io/cloudevents/extensions/ExtensionsParser.java
slinkydeveloper: https://github.com/cloudevents/sdk-java/issues/122

slinkydeveloper:
@bsideup Everything that relates to extensions read/write for now is just done to be the bare minimum compilable, I've opened this issue to discuss apis about it https://github.com/cloudevents/sdk-java/issues/122

api/src/main/java/io/cloudevents/Attributes.java
slinkydeveloper: Opened this one to discuss about it https://github.com/cloudevents/sdk-java/issues/123

slinkydeveloper:
I removed the WIP flag, for me this PR is ready to merge.

fabiojose:
LGTM

PR #104, opened 2020-03-19
Fix content-length

matejvasek:
Fixes #103

PR #129, opened 2020-04-23
Reworked Vert.x Integration

### LINKED ISSUE #112: Modularization: http-vertx module
This module implements the [HTTP Protocol Binding](https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md) with vertx.

Actually, we have this module, but it needs refactorings.

- `http-vertx`

__This module depend on [api](https://github.com/cloudevents/sdk-java/issues/111) one.__

When someone wants the http binding to create or read events with vertx, they add this module and they concern about potential conflicts with pre-existing dependencies in their projects.

slinkydeveloper:
Fix #112

Supersede #107, #104, #116


matejvasek:
While testing I found out that headers are case sensitive. For instance if I `curl` with `-H "Ce-Id: 42"` the id is not read. Maybe we should pre-process headers with `toLowerCase()`.

slinkydeveloper:
That's weird, `MultiMap#get` should be case-insensitive

slinkydeveloper:
@matejvasek can you post the test code you wrote?

http/vertx/src/main/java/io/cloudevents/http/vertx/impl/BinaryVertxMessageImpl.java
```diff
@@ -0,0 +1,84 @@
...
+        return Encoding.BINARY;
+    }
+
+    @Override
+    public <T extends BinaryMessageVisitor<V>, V> V visit(BinaryMessageVisitorFactory<T, V> visitorFactory) throws MessageVisitException, IllegalStateException {
+        BinaryMessageVisitor<V> visitor = visitorFactory.createBinaryMessageVisitor(this.version);
+
+        // Grab from headers the attributes and extensions
+        this.headers.forEach(e -> {
+            if (e.getKey().startsWith(CE_PREFIX)) {
```
matejvasek, line 54: Shouldn't be toLowerCase used here?

slinkydeveloper:
@matejvasek fixed, now it works :smile:

PR #134, opened 2020-04-27
Reworked Kafka module

### LINKED ISSUE #114: Modularization: kafka module
This module implements the [Kafka Protocol Binding](https://github.com/cloudevents/spec/blob/v1.0/kafka-protocol-binding.md)

This module already exists, but it needs refactorings.

- `kafka`

__This module depend on [api](https://github.com/cloudevents/sdk-java/issues/111) one.__

When someone wants the kafka binding to create or read events, they add this module and they concern about potential conflicts with pre-existing dependencies in their projects.

slinkydeveloper:
Fix #114

Now the `Kafka` module doesn't provide any more a consumer/producer, but it provides:

* `KafkaMessageFactory` and `KafkaProducerMessageVisitor` to manually serialize/deserialize
* `Serializer` and `Deserializer` implementations to use with Kafka APIs


api/src/main/java/io/cloudevents/message/impl/BaseGenericBinaryMessageImpl.java
```diff
@@ -0,0 +1,71 @@
...
+
+        // Set the payload
+        if (this.body != null && this.body.length != 0) {
+            visitor.setBody(this.body);
+        }
+
+        return visitor.end();
+    }
+
+    protected abstract String getContentTypeHeaderKey();
```
bsideup, line 63: Could use type parameterization for key's type

api/src/main/java/io/cloudevents/message/impl/BaseGenericBinaryMessageImpl.java
```diff
@@ -0,0 +1,71 @@
...
+                    if (this.version.getAllAttributes().contains(name)) {
+                        visitor.setAttribute(name, headerValueToString(v));
+                    } else {
+                        visitor.setExtension(name, headerValueToString(v));
+                    }
+                }
+                if (k.equalsIgnoreCase(getContentTypeHeaderKey())) {
+                    visitor.setAttribute("datacontenttype", headerValueToString(v));
+                }
+            } catch (StringIndexOutOfBoundsException ex) {
```
bsideup, line 50: You could also use `k.startsWith` to avoid this exception

api/src/main/java/io/cloudevents/message/impl/BaseGenericBinaryMessageImpl.java
```diff
@@ -0,0 +1,71 @@
...
+        this.version = version;
+        this.body = body;
+    }
+
+    @Override
+    public <T extends BinaryMessageVisitor<V>, V> V visit(BinaryMessageVisitorFactory<T, V> visitorFactory) throws MessageVisitException, IllegalStateException {
+        BinaryMessageVisitor<V> visitor = visitorFactory.createBinaryMessageVisitor(this.version);
+
+        // Grab from headers the attributes and extensions
+        this.forEachHeader((k, v) -> {
```
bsideup, line 34: can it be `key` and `value`?

api/src/main/java/io/cloudevents/message/impl/BaseGenericBinaryMessageImpl.java
```diff
@@ -0,0 +1,71 @@
...
+                    if (name.equals("specversion")) {
+                        return;
+                    }
+                    if (this.version.getAllAttributes().contains(name)) {
+                        visitor.setAttribute(name, headerValueToString(v));
+                    } else {
+                        visitor.setExtension(name, headerValueToString(v));
+                    }
+                }
+                if (k.equalsIgnoreCase(getContentTypeHeaderKey())) {
```
bsideup, line 47:
```suggestion
                else if (k.equalsIgnoreCase(getContentTypeHeaderKey())) {
```

api/src/main/java/io/cloudevents/message/impl/BaseGenericBinaryMessageImpl.java
slinkydeveloper: No we can't because in any case the **key** must be string, because during the header processing the method `equalsIgnoreCase` is used.

api/src/main/java/io/cloudevents/message/impl/BaseGenericBinaryMessageImpl.java
slinkydeveloper: `startsWith` is not case insensitive

bsideup: then I'd suggest checking the length. Exceptions in Java are expensive.

api/src/main/java/io/cloudevents/message/impl/BaseGenericBinaryMessageImpl.java
bsideup: Then you can use `protected abstract boolean isContentTypeHeader(K key)`, especially given that some frameworks already provide case-insensitive headers

api/src/main/java/io/cloudevents/message/impl/BaseGenericBinaryMessageImpl.java
bsideup: @slinkydeveloper I will appreciate if you let me resolve the conversations I started. Not really sure why this one is marked as resolved

slinkydeveloper:
@bsideup can you review it now?

api/src/main/java/io/cloudevents/message/impl/BaseGenericBinaryMessageImpl.java
```diff
@@ -0,0 +1,69 @@
...
+        if (this.body != null && this.body.length != 0) {
+            visitor.setBody(this.body);
+        }
+
+        return visitor.end();
+    }
+
+    protected abstract boolean isContentTypeHeader(HK key);
+
+    protected abstract boolean isCEPrefixed(HK key);
```
bsideup, line 61: Perhaps `isCloudEventsHeader` would be better, to not expose the implementation (some CE-native systems may not have the `ce-` prefix)

api/src/main/java/io/cloudevents/message/impl/BaseGenericBinaryMessageImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+    protected abstract String stripKeyPrefixAndParse(HK key);
```
bsideup, line 63: same here, something like `toCloudEventsHeader` would be better (the case should match the spec - currently lower-cased)

api/src/main/java/io/cloudevents/message/impl/BaseGenericBinaryMessageImpl.java
```diff
@@ -0,0 +1,69 @@
...
+                String name = stripKeyPrefixAndParse(key).toLowerCase();
+                if (name.equals("specversion")) {
+                    return;
+                }
+                if (this.version.getAllAttributes().contains(name)) {
+                    visitor.setAttribute(name, headerValueToString(value));
+                } else {
+                    visitor.setExtension(name, headerValueToString(value));
+                }
+            } else if (isContentTypeHeader(key)) {
```
bsideup, line 46: I'd revert the `if-else`. Otherwise, `ce-content-type` (let's assume some system uses it) would go into the first branch.

api/src/main/java/io/cloudevents/message/impl/BaseGenericBinaryMessageImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+    protected abstract boolean isContentTypeHeader(HK key);
+
+    protected abstract boolean isCEPrefixed(HK key);
+
+    protected abstract String stripKeyPrefixAndParse(HK key);
+
+    protected abstract void forEachHeader(BiConsumer<HK, HV> fn);
+
+    protected abstract String headerValueToString(HV value);
```
bsideup, line 67: `toString(HV value)` would work, too. The argument's type defines the scope (header's value) anyways

slinkydeveloper: I would prefer to have consistency with the key converter method, look at it now.

kafka/README.md
```diff
@@ -1,220 +1,91 @@
...
-
-Add the dependency in your project:
+For Maven based projects, use the following to configure the [Kafka Protocol Binding](https://github.com/cloudevents/spec/blob/master/kafka-protocol-binding.md):
 
 ```xml
 <dependency>
     <groupId>io.cloudevents</groupId>
-    <artifactId>cloudevents-kafka</artifactId>
-    <version>1.3.0</version>
+    <artifactId>http-vertx</artifactId>
```
bsideup, line 8: uh? Shouldn't it be `cloudevents-kafka`?

kafka/pom.xml
```diff
@@ -7,117 +7,67 @@
...
-  </properties>
+    <dependencies>
+        <dependency>
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
```
bsideup, line 36: provided scope?

kafka/README.md
slinkydeveloper: wtf something went wrong with the rebase

kafka/src/main/java/io/cloudevents/kafka/KafkaMessageFactory.java
```diff
@@ -0,0 +1,28 @@
...
+import io.cloudevents.kafka.impl.KafkaBinaryMessageImpl;
+import io.cloudevents.kafka.impl.KafkaHeaders;
+import io.cloudevents.message.Message;
+import io.cloudevents.message.impl.GenericStructuredMessage;
+import io.cloudevents.message.impl.MessageUtils;
+import io.cloudevents.message.impl.UnknownEncodingMessage;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.common.header.Headers;
+
+public interface KafkaMessageFactory {
```
bsideup, line 12: why interface? Consider using a final class with private constructor

kafka/pom.xml
slinkydeveloper: Mh I would prefer to have it compile with compile scope

kafka/src/main/java/io/cloudevents/kafka/impl/KafkaHeaders.java
```diff
@@ -0,0 +1,63 @@
...
+ * @author Francesco Guardiani
+ * @author Florian Hussonnois
+ */
+public class KafkaHeaders {
+
+    public static final String CE_PREFIX = "ce_";
+
+    public static final String CONTENT_TYPE = "content-type";
+
+    public static final Map<String, String> ATTRIBUTES_TO_HEADERS = Stream.concat(
```
bsideup, line 40: is this Map really necessary? What does it solve?

kafka/README.md
```diff
@@ -1,220 +1,91 @@
...
-
-Add the dependency in your project:
+For Maven based projects, use the following to configure the [Kafka Protocol Binding](https://github.com/cloudevents/spec/blob/master/kafka-protocol-binding.md):
 
 ```xml
 <dependency>
     <groupId>io.cloudevents</groupId>
     <artifactId>cloudevents-kafka</artifactId>
-    <version>1.3.0</version>
+    <version>2.0.0-SNAPSHOT </version>
```
bsideup, line 9:
```suggestion
    <version>2.0.0-SNAPSHOT</version>
```

kafka/src/main/java/io/cloudevents/kafka/impl/KafkaHeaders.java
slinkydeveloper: It pre-caches the attribute to headers name, avoiding to allocate a new one every time.

slinkydeveloper: and it can be definitely protected

bsideup: Since it requires a code change every time a new SpecVersion is released, I'd suggest to not introduce this optimization yet, unless someone complains about this short-living strings allocated. Just my 2c.

slinkydeveloper:
I go ahead and merge it, eventual follow-up will fix the missing bits

PR #138, opened 2020-04-28
Introduce JSR-305 compatible nullability, remove `Optional`

### LINKED ISSUE #123: Returning empty values
From https://github.com/cloudevents/sdk-java/issues/119 and https://github.com/cloudevents/sdk-java/pull/108#discussion_r411193730:

What technique should we use, around the sdk, to notify the user of an eventual empty return value?

1. `Optional` getters
2. Have an annotation called `@Nullable` which notifies the user of an eventual null return value
3. Just explicit in the Javadoc that a method can return null

bsideup:
Closes #123.

Signed-off-by: Sergei Egorov <segorov@pivotal.io>

### DIFF
--- a/api/src/main/java/io/cloudevents/lang/Nullable.java
+++ b/api/src/main/java/io/cloudevents/lang/Nullable.java
@@ -1,0 +1,19 @@
+package io.cloudevents.lang;
+
+import javax.annotation.Nonnull;
+import javax.annotation.meta.TypeQualifierNickname;
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.*;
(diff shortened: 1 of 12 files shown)


api/src/main/java/io/cloudevents/lang/Nullable.java
```diff
@@ -0,0 +1,19 @@
+package io.cloudevents.lang;
```
olegz, line 1: Should we include license header?

slinkydeveloper: Since most license headers are broken, let's do it in a followup with formatting

slinkydeveloper:
@bsideup can i merge it?

bsideup:
@slinkydeveloper yes, please 👍 Thanks for the quick review!

PR #140, opened 2020-04-28
Formatted code and added copyright headers

slinkydeveloper:

### DIFF
--- a/.travis.settings.xml
+++ b/.travis.settings.xml
@@ -1,5 +1,5 @@
 <!--
-  ~ Copyright 2020 The CloudEvents Authors
+  ~ Copyright 2018-Present The CloudEvents Authors
   ~ <p>
   ~ Licensed under the Apache License, Version 2.0 (the "License");
   ~ you may not use this file except in compliance with the License.
--- a/.travis.settings.xml
+++ b/.travis.settings.xml
@@ -1,3 +1,20 @@
+<!--
+  ~ Copyright 2020 The CloudEvents Authors
+  ~ <p>
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~ <p>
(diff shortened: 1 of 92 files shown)


.travis.settings.xml
```diff
@@ -1,3 +1,20 @@
+<!--
+  ~ Copyright 2020 The CloudEvents Authors
```
bsideup, line 2: Consider using `2018-Present` (legally valid form) to avoid having to bump it every year

slinkydeveloper: Mh i felt the same, let's see what @duglin thinks about it

slinkydeveloper: Also the same should apply here i suppose https://github.com/cloudevents/sdk-java/pull/140/files#diff-9879d6db96fd29134fc802214163b95aR189

bsideup: context: Spring projects have recently migrated to `XXXX-Present` scheme after a thoughtful review by the legal team to avoid having to deal with the year bumps that create a ton of noise in work in projects every year.

bsideup: > Also the same should apply here i suppose

I'd say... everywhere :) Bumps are PITA and we have a chance to avoid it early :)

duglin: if some legal person someplace says it's "ok", then it sounds good to me :-)

bsideup: I should have probably mentioned that I am not a layer (but got advised by them :D)

https://github.com/reactor/reactor-core/blob/a1dab8ad71605bfcf5e0d185ebe6e0f9f5ef560d/buildSrc/build.gradle#L2

slinkydeveloper: @duglin i thought you're the legal guy here :smile:

duglin: I think I should be insulted by that LOL

slinkydeveloper:
Ok I changed all copyright years with 2018-Present

PR #151, opened 2020-05-17
Implement dataref extension

paulschwarz:
Implements dataref extension (see #70) targeting SDK version 2 (see #71)

### DIFF
--- a/core/src/main/java/io/cloudevents/core/extensions/DatarefExtension.java
+++ b/core/src/main/java/io/cloudevents/core/extensions/DatarefExtension.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2018-Present The CloudEvents Authors
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


paulschwarz:
Implementation is based on the DistributedTracingExtension, however, I haven't been able to test this end to end to verify that it really works as expected because I can't see how to use the version 2 SDK properly. Everything I've tried results in unresolved classes.

slinkydeveloper:
@paulschwarz I think the two tests you included are enough.

> I can't see how to use the version 2 SDK properly. Everything I've tried results in unresolved classes.

Can you be more specific about that?

api/src/main/java/io/cloudevents/extensions/DatarefExtension.java
```diff
@@ -0,0 +1,86 @@
...
+package io.cloudevents.extensions;
+
+import io.cloudevents.CloudEvent;
+import io.cloudevents.Extension;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public final class DatarefExtension implements Extension {
```
slinkydeveloper, line 27: Can you add a javadoc with a link to the extension spec?

api/src/main/java/io/cloudevents/extensions/DatarefExtension.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public static final String DATAREF = "dataref";
+
+    private String dataref;
```
slinkydeveloper, line 31: Since dataref from the spec is a [`uri-reference`](https://github.com/cloudevents/spec/blob/master/extensions/dataref.md#dataref), i think we should use as type `URI`

slinkydeveloper:
Ok so when you rebase, please move this code in the core module (api was renamed to core and api is the new only interfaces module)

slinkydeveloper:
Also, `Extension` interface was changed a bit (nothing to be worried about), can you rebase on it please?

PR #150, opened 2020-05-15
Re: Proposal for reworking the unstructured read/write of events

slinkydeveloper:
This proposal comes from discussions in #143 and https://docs.google.com/document/d/1VY85Afhb-oj7m9Sm84U_f8kJp6boBS8nnuTMc9tkgO8/edit?usp=sharing

The goal of these changes is to provide an unstructured view of the event and an unstructured manner to write the event, without giving additional burden to the `CloudEvent` implementer (like `asStructuredMessage()` and `asBinaryMessage()` does). It's not a goal of this PR to address the other concerns of #143.

## `CloudEventVisitable`

`CloudEvent` implements `CloudEventVisitable`, which is an interface that defines how an unstructured visit of the message happens. `CloudEvent` provides a default implementation for it, avoiding to put this burden to the `CloudEvent` implementer.

`CloudEventVisitable` is used by `EventFormat` implementations and protocol binding implementations, keeping low the allocation cost for encoding/decoding an event.

Now users can write binary messages using:

```
event
  .visit(VertxHttpServerResponseMessageVisitor.create(req.response()));
```

While they can write structured message using:

```
GenericStructuredMessage.fromEvent(event)
  .visit(VertxHttpServerResponseMessageVisitor.create(req.response()));
```

Two helper methods are also provided for that.

## `CloudEventBuilder` and `CloudEventVisitor`/`CloudEventVisitorFactory`

`CloudEventBuilder` is the "write side" of `CloudEvent`. Concrete implementations provides "structured write" methods to write the `CloudEvent` (`withId`, `withSource`, etc) and implements the `CloudEventVisitor<CloudEvent>` interface, which provides "unstructured write" methods to create an event starting from a binary message.

## Event conversion

The visitAttributes will be used as basis to reimplement specversion conversion, in order to remove the methods on the CloudEvent interface. This is shown in #147

Signed-off-by: Francesco Guardiani <francescoguard@gmail.com>

PR #148, opened 2020-05-15
Flatten attributes interface/implementation

slinkydeveloper:
This PR proposes to flatten the `Attributes` interface (renamed `CloudEventAttributes`) to `CloudEvent`

Depends on #147

Issue #118, opened 2020-04-20
Remove `Attributes` interface


bsideup:
In the current version of SDK, one has to deal with generic signature of `CloudEvent` and an unnecessary allocation of the `AttributesImpl` object.

While working on [CloudEvents for Liiklus](https://github.com/bsideup/liiklus/pull/252/files#diff-7bb8ccda936f38f78c9d74c4430b107cR190), I had to implement our own `CloudEvent` type that implements both `CloudEvent` and `Attributes` to reduce the allocations (see the link)

Since (according to the spec) there is always an 1:1 mapping of an event and its attributes, I suggest to remove the interface and make the fields part of the `CloudEvent` interface.

To make the change less impactful and support a migration path, I suggest we do it in two steps (that should be released separately):

1. Make `CloudEvent` implementations implement both `CloudEvent` and `Attributes` and return `this` from `getAttributes()`
2. Make `CloudEvent` extend from `Attributes` and turn `getAttributes()` into a default method that returns `this` and deprecate `getAttributes()` and `Attributes`
3. Remove `Attributes` and ``getAttributes()` and move the fields to the `CloudEvent` interface

If the community agrees, I can work on all 3 PRs.

slinkydeveloper:
Now that the generics problem doesn't exist anymore, I think this discussion really brings us down to the "inheritance vs composition" debate, which doesn't have a real winner here. I think we should keep as is for several reasons:

* Easier to implement conversions back/forth different spec versions
* Easier to implement `Message` interfaces
* Less burden to have a new spec version
* Only one `CloudEvent` interface impl that already handles all the `CloudEvent` various methods
* A user can handle `Attributes` separately to the `Event` and eventually cast down the specific `AttributesImpl` to handle specversion specific informations, and we can still hide the details of `CloudEventImpl`
* Most important IMO: consistency with other SDKs

From a usability POV, I think `CloudEvent` interface can extend `Attributes` and delegate the methods to `AttributesImpl`.

PR #147, opened 2020-05-15
Reimplemented Spec version conversion

slinkydeveloper:
Depends on #150

Removed spec version conversion inside `CloudEvent` interface and `Attributes` interface. Now to convert an event from one spec version to another, user should go through builders:

```java
CloudEvent eventV1 = CloudEventBuilder.v1(eventV03).build();
```

Signed-off-by: Francesco Guardiani <francescoguard@gmail.com>

PR #155, opened 2020-05-21
Proposed set of APIs

slinkydeveloper:
Note: the first commit moves cloudevents-api to a new module called cloudevents-core, while cloudevents-api effectively contains the proposed apis

The second commit effectively shows the APIs only package.

Closes #143, #118 and supersedes #148. Takes some bits from #150 but most of that needs to be rebased after this one goes in.

This doesn't affect the existing implementation of CloudEvent/CloudEventBuilder/protocol bindings/event formats etc, I'll *personally* take care of re-glueing things together with cloudevents-core as soon as we get this one in.


api/src/main/java/io/cloudevents/CloudEvent.java
```diff
@@ -31,47 +28,11 @@
  * @author slinkydeveloper
  */
 @ParametersAreNonnullByDefault
-public interface CloudEvent {
-
-    /**
-     * The event context attributes
-     */
-    Attributes getAttributes();
+public interface CloudEvent extends CloudEventAttributes, CloudEventExtensions {
```
n3wscott, line 31: this is nice and clean, a cloudevent is attributes, extensions and data.

duglin:
@slinkydeveloper thanks for proposing this!

api/src/main/java/io/cloudevents/CloudEventAttributes.java
```diff
@@ -0,0 +1,86 @@
...
+
+import javax.annotation.ParametersAreNonnullByDefault;
+import java.net.URI;
+import java.time.ZonedDateTime;
+import java.util.Set;
+
+/**
+ * The marker interface for CloudEvents attributes
+ *
+ * @author fabiojose
```
olegz, line 30: What is the policy around author tags. I see we have some lingering names but clearly fabio did not author this class. Just curious

api/src/main/java/io/cloudevents/visitor/CloudEventAttributesVisitor.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package io.cloudevents.visitor;
+
+import io.cloudevents.types.Time;
+
+import java.net.URI;
+import java.time.ZonedDateTime;
+
+@FunctionalInterface
```
olegz, line 25: Please remove that as it is not a functional interface

api/src/main/java/io/cloudevents/visitor/CloudEventExtensionsVisitor.java
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.cloudevents.visitor;
+
+@FunctionalInterface
```
olegz, line 20: Same as previous comment. Basically we need to ensure that we don't throw `@FunctionalInterface` around. It has a clear meaning.

core/src/main/java/io/cloudevents/CloudEvent.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Map;
+
+/**
+ * An abstract event envelope
+ *
+ * @author fabiojose
+ * @author slinkydeveloper
+ */
+@ParametersAreNonnullByDefault
+public interface CloudEvent {
```
olegz, line 34: This needs to be removed

api/src/main/java/io/cloudevents/CloudEventAttributes.java
```diff
@@ -0,0 +1,86 @@
...
+
+import io.cloudevents.lang.Nullable;
+
+import javax.annotation.ParametersAreNonnullByDefault;
+import java.net.URI;
+import java.time.ZonedDateTime;
+import java.util.Set;
+
+/**
+ * The marker interface for CloudEvents attributes
```
olegz, line 28: We should rephrase it as it is clearly not a marker interface. Consider the following:
```
Strategy which defines CloudEvent attributes as per specification - https://github.com/cloudevents/spec/blob/v1.0/spec.md#context-attributes
```

olegz:
I am in favor of this PR providing the review comments are addressed.
- The core concerns expressed in #143 are addressed in this PR, so I am rejecting #143 with comments clarifying that this PR effectively superseded #143 by accepting and including the changes proposed by #143. 
- There are no more dependency of `CloudEvent` on `CloudEvent*Visitor(s)` (the concern expressed in #146) making the newly defined visitor strategies optional.
- The #146 should be rejected as well, since it is no longer valid in lieu of this PR

`CloudEvent` is now clean, concise and free of any assumptions.
Also, by defining a separate `-api` module, it allows different implementations of CloudEvent to co-exist.

core/src/main/java/io/cloudevents/SpecVersion.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    public static SpecVersion parse(String sv) {
+        switch (sv) {
+            case "0.3":
+                return SpecVersion.V03;
+            case "1.0":
+                return SpecVersion.V1;
+            default:
+                throw new IllegalArgumentException("Unrecognized SpecVersion " + sv);
```
bsideup, line 65: Just an idea:
What if instead of throwing we return `SpecVersion.UNKNOWN` (does not allow any attribute)?
This way, it will be a little bit easier to deal with this method from functional code, plus there will be no need to know which exact exception type is thrown.

If not, we should at leat add `@throws` to method's javadoc (or maybe introduce a custom checked type then)

core/src/main/java/io/cloudevents/lang/Nullable.java
```diff
@@ -0,0 +1,36 @@
...
+import static java.lang.annotation.ElementType.*;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+import static javax.annotation.meta.When.MAYBE;
+
+@Target(value = {METHOD, PARAMETER, FIELD})
+@Retention(value = RUNTIME)
+@Documented
+@Nonnull(when = MAYBE)
+@TypeQualifierNickname
+public @interface Nullable {
```
bsideup, line 35: This file belongs to `-api`

api/src/main/java/io/cloudevents/CloudEventAttributes.java
slinkydeveloper: not sure we have a policy about tags, for now i just remove it since it's clearly incorrect

api/src/main/java/io/cloudevents/CloudEventAttributes.java
slinkydeveloper: I don't wanna touch any javadocs in this pr, i'm planning to do it in future as soon as we have other things aligned

slinkydeveloper: Most of the javadocs here are wrong or missing, so let's just focus on the interface now

slinkydeveloper:
After merging this one i'm gonna merge the #150, #147 and #148 because i made some progress on the implementations, but i *won't* use the interfaces introduced by them. I'm gonna glue back the progress made there with these new interfaces in new PRs

core/src/main/java/io/cloudevents/lang/Nullable.java
slinkydeveloper: yes but i still didn't introduced in this pr the dependency api -> core, so for now i need it there otherwise the code doesn't compile

api/src/main/java/io/cloudevents/visitor/CloudEventAttributesVisitor.java
```diff
@@ -0,0 +1,60 @@
...
+public interface CloudEventAttributesVisitor {
```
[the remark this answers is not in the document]
slinkydeveloper: No, the visitor won't change. It is designed to be forward and backward compatible and a good implementation of CloudEvent must implement it in order to access to the other sdk features. So this really belongs to api and not to core

core/src/main/java/io/cloudevents/SpecVersion.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public enum SpecVersion {
+    V03(
+        "0.3",
+        Collections.unmodifiableSet(new HashSet<>(Arrays.asList("specversion", "id", "type", "source"))),
```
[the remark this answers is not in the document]
slinkydeveloper: Why? Is there any difference?

core/src/main/java/io/cloudevents/SpecVersion.java
```diff
@@ -0,0 +1,89 @@
...
+package io.cloudevents;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public enum SpecVersion {
```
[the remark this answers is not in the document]
slinkydeveloper: https://github.com/cloudevents/sdk-java/pull/155#discussion_r429086013

api/src/main/java/io/cloudevents/CloudEventBuilder.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.cloudevents;
+
+public interface CloudEventBuilder<CE extends CloudEvent> {
```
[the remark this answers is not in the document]
slinkydeveloper: we can put that back in the core if people wants to, to me it's even better to have it on the core so i can have the nice static methods `CloudEventBuilder.v1()`...

api/pom.xml
```diff
@@ -15,15 +15,16 @@
...
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
-    <modelVersion>4.0.0</modelVersion>
 
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <parent>
-        <groupId>io.cloudevents</groupId>
         <artifactId>cloudevents-parent</artifactId>
+        <groupId>io.cloudevents</groupId>
```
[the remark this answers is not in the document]
slinkydeveloper: ?

core/src/main/java/io/cloudevents/SpecVersion.java
slinkydeveloper: I don't think it's a good idea, I'm not prone adding a new variant to the enum for "UNKNOWN", only because we don't want to fail the parsing process with an exception. I could rather return a null if it's not found, but adding an enum variant it's not correct imo.

About the javadoc, i didn't touched any of them https://github.com/cloudevents/sdk-java/pull/155#discussion_r429084873

api/src/main/java/io/cloudevents/CloudEventAttributes.java
olegz: I understand your concern about some lingering javadocs, but this javadoc is clearly a new javadoc created as part of this PR so let's fix it here is well. 
No need to deliberately introduce something new that is already acknowledged by both of us as "wrong" with expectations that it is going to get fixed some time in the future.

slinkydeveloper: Now it should be fine

slinkydeveloper: let me know if it needs more refinement

olegz: all good

core/src/main/java/io/cloudevents/SpecVersion.java
slinkydeveloper: Ok i rewrote the javadocs for this method

slinkydeveloper:
Ok I can't say why but github doesn't allow me to re-request again a review, @olegz @bsideup can you review this pr again?

api/src/main/java/io/cloudevents/visitor/CloudEventAttributesVisitor.java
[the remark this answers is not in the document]
olegz: My opinion is that visitor is just one mechanism which may or may not be the best for a given task. And yes I agree with Sergei that it does not belong here, but given that the real value of this PR is the fact that it accepts the changes proposed by #143 (streamlining CloudEvent interface), i say we can debate the placement of the visitor interface some other time.

core/src/main/java/io/cloudevents/lang/Nullable.java
bsideup: It really should be `core -> api`, not the other way around.

slinkydeveloper: Oh yes i'm sorry, i mistyped :smile:

core/src/main/java/io/cloudevents/SpecVersion.java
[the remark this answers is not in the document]
slinkydeveloper: Ah i see now, this is not possible because i need to distinguish between mandatory and optional attributes, so varargs doesn't fit here

api/src/main/java/io/cloudevents/CloudEventBuilder.java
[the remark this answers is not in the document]
slinkydeveloper: I removed it, i'll put it on core

api/src/main/java/io/cloudevents/visitor/CloudEventAttributesVisitor.java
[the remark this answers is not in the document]
bsideup: @slinkydeveloper 

>  It is designed to be forward and backward compatible and a good implementation of CloudEvent must implement it in order to access to the other sdk features

After looking at `-api`, I got a feeling that I can implement everything just by using `CloudEvents` interface. That makes me think that the visitors are optional, or even a detail of the implementation of `core`.

What I suggest here is to put them into `core`, and when/if we feel that they make sense exposing publicly as part of the API, we move them there as a follow up. WDYT? :)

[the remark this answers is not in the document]
slinkydeveloper: I think it's not a good idea, as i clearly underlined in my other proposals and yesterday meeting.

api/pom.xml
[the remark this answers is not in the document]
bsideup: just the ordering thing, not critical :)

core/src/main/java/io/cloudevents/SpecVersion.java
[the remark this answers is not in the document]
bsideup: ah, okay, I see. Well, maybe we can at least do `Collections.unmodifiableSet(new HashSet<>(` in the constructor

api/src/main/java/io/cloudevents/visitor/CloudEventAttributesVisitor.java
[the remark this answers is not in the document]
slinkydeveloper: If in the process of vendoring the sdk you think these interfaces are useless, feel free to remove them from the classpath.

api/src/main/java/io/cloudevents/CloudEventBuilder.java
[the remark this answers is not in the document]
bsideup: Cool, thanks! 👍

api/pom.xml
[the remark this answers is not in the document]
slinkydeveloper: Ah ok i didn't get it, lemme fix it

core/src/main/java/io/cloudevents/SpecVersion.java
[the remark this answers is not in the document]
slinkydeveloper: Sure!

api/src/main/java/io/cloudevents/visitor/CloudEventAttributesVisitor.java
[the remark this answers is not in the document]
bsideup: > If in the process of vendoring the sdk you think these interfaces are useless, feel free to remove them from the classpath using one of the various maven plugins designed for this purpose.

that's a dependency, one cannot alter dependencies or anything.

But okay, just wanted to state that these visitors are useless for those who only want to implement `CloudEvents` interface (e.g. by wrapping an existing type), so it really becomes a detail of the implementation of `core`.

Since there are no references to the visitors in `CloudEvents` interface, It does not block me, I can just choose not to provide any implementation for them 👍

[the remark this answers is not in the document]
slinkydeveloper: You can always vendor a dependency shipping these interfaces with your jar (eg spring-cloudevents vendors directly cloudevents-api shipping a jar with these interfaces inside its classpath). In that case, you can remove classpath elements using the maven-jar-plugin.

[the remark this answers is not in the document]
bsideup: @slinkydeveloper we don't do this in Java. This will also break classpaths of some projects.

slinkydeveloper:
> will break some systems 

Production systems should not depend on a SNAPSHOT in first instance :smile:

bsideup:
@slinkydeveloper true :D Although I believe this project a very young and atm we should think about greenfield / experimental projects (including your knative vertx sample that uses SNAPSHOT and will break, btw), not production ones.

olegz:
While formally registering the same concern about visitor interface to be removed from the API module (mentioned by Sergei), I also have no objection for merging this PR.
+1

n3wscott:
LGTM

Issue #122, opened 2020-04-21
Read and write extensions APIs


slinkydeveloper:
Define APIs about how to read and write extensions

slinkydeveloper:
#108 provides this to read extensions https://github.com/cloudevents/sdk-java/blob/ae4028d59d92cfa162f67048c228f546c1e94dde/api/src/main/java/io/cloudevents/extensions/ExtensionsParser.java

This way the user can read extensions like:

```
DistributedTracingExtension ext = ExtensionsParser.getInstance().parseExtension(DistributedTracingExtension.class, event);
```

While to write extensions, the `CloudEventBuilder` supports them:

```
builder.withExtension(ext);
```

slinkydeveloper:
`Extension` needs to be reworked to remove the assertion that it can be translated to a map (aka similar to `CloudEventExtensions`)

PR #157, opened 2020-05-29
Glue core & api

### LINKED ISSUE #156: Plan for glue-ing core to api
1. [ ] Remove `toBinaryMessage` & `toStructuredMessage` implementing the CloudEventVisitor/CloudEventVisitable:  #150. This allows all protocol bindings/event formats to continue working
2. [ ] Remove spec version conversion (`toV1()` and `toV03()`) from `CloudEvent` interface: https://github.com/cloudevents/sdk-java/pull/147
3. [ ] Flatten attributes implementation (this is not required, however it simplifies the code and reduce allocations): WIP from https://github.com/cloudevents/sdk-java/pull/148
4. [ ] Remove assumption that `CloudEvent` implements `CloudEventVisitable`: This PR will move the code under the default implementation of `CloudEventVisitable` under a generic `CloudEventVisitable` implementation for `CloudEvent` interface (placed in `cloudevents-core`)
5. [ ] `cloudevents-core` depends on `cloudevents-api`

All PRs from 1 to 4 will change the `CloudEvent` interface (and friends) inside **core** in order to make it equal to the one inside **api**. After this is done, it will be trivial to remove the old interfaces in **core** and replace them with the ones in **api**.

slinkydeveloper:
Fix #156

Each commit represent one of the steps explained in #156.

Deprecates #147 and #150


core/src/main/java/io/cloudevents/builder/CloudEventBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.cloudevents.builder;
```
bsideup, line 18: In lue of Java Module System that does not allow the same package to be defined in multiple modules, I'd recommend to put everything in `core` to `io.cloudevents.impl` sub-packages. So that, if we ever decide to expose builders in the API, we will no have a problem of `io.cloudevents.builder` and similar being defined in `core` and not `api`

core/src/main/java/io/cloudevents/message/Message.java
```diff
@@ -18,42 +18,90 @@
 package io.cloudevents.message;
```
bsideup, line 18: see https://github.com/cloudevents/sdk-java/pull/157/files#r432390527

core/src/main/java/io/cloudevents/v03/CloudEventBuilder.java
```diff
@@ -16,24 +16,23 @@
  */
 package io.cloudevents.v03;
```
bsideup, line 17: see https://github.com/cloudevents/sdk-java/pull/157/files#r432390527

core/src/test/java/io/cloudevents/extensions/DistributedTracingExtensionTest.java
```diff
@@ -17,6 +17,7 @@
 package io.cloudevents.extensions;
```
bsideup, line 17: see https://github.com/cloudevents/sdk-java/pull/157/files#r432390527

bsideup:
@slinkydeveloper nice! Happy to see the `api` module untouched 👍

core/src/main/java/io/cloudevents/builder/CloudEventBuilder.java
slinkydeveloper: ok I'm sorry but you find me a little unprepared on this. Why does the java module system should break? There are no packages that are defined both in core and api at the moment.

And any reason why `impl` more than `core` (so it reflects the module name)?

slinkydeveloper: api defines:

- `io.cloudevents`
- `io.cloudevents.lang`
- `io.cloudevents.types`
- `io.cloudevents.visitor`

while core defines:

- `io.cloudevents.impl`
- `io.cloudevents.builder`
- `io.cloudevents.message`
- `io.cloudevents.format`
- `io.cloudevents.extensions`
- `io.cloudevents.v1`
- `io.cloudevents.v03`

slinkydeveloper: If for some reason someday we want to move `message` or `builder` to api we move the whole package, i carefully avoided to put both in `message` and `builder` only the interfaces and not the implementations

slinkydeveloper: Does that makes sense?

bsideup: Yes, sorry, I wasn't clear enough. Currently, there is no issue indeed.

But, let's say one day we decide to add `extensions` sub-package to `api`.
We won't be able to use `io.cloudevents.extensions` because it is already defined in `core`, so we would need to find another (probably hacky) name for it.

If we put everything in `core` into `io.cloudevents.impl.*`, we protected future ourselves from having to fight with JPMS.
As a neat side effect, It also makes it clear that any import from `io.cloudevents.impl.*` is the implementation detail.

bsideup: > we move the whole package

this may break projects that have newer `api.jar` but older `core.jar` (can easily happen when a dependency defines only `api` as its dependency, and uses a newer one)

slinkydeveloper: Ok but if we ever move a package to api, it's an _"api"_ package, so it won't contain any implementation (in theory) and you want to move the whole package (excluding subpackages), not just a couple of interfaces splitting the package between api and core. For the particular case of `message` and `builder`, I don't think splitting it moving a part to `api` and keeping a part in `core` makes sense. Do you foresee one particular module that could require such "splitting"?

The case of `extensions` it's interesting because extensions IMO should contain only the "extension pojos", but now it contains that `ExtensionsParser` that probably needs to be moved somewhere else. For that particular case I've opened the issue #122 to discuss about Extensions. I think the same issue applies to `format` package.

> As a neat side effect, It also makes it clear that any import from io.cloudevents.impl.* is the implementation detail.

Yeah but it has some interfaces too that other modules uses, that's why I prefer `core`.

slinkydeveloper: > this may break projects that have newer api.jar but older core.jar (can easily happen when a dependency defines only api as its dependency, and uses a newer one)

Ok but this is the kind of change done between majors, but yes i see the point

slinkydeveloper: let me try to move to `io.cloudevents.core`

bsideup: > Yeah but it has some interfaces too that other modules uses

Ideally, the modules should also define their own unique packages.

Just sharing some **very** painful experience working on a library that did not expect JPMS and such limitations to happen 😂

slinkydeveloper: Check it now

core/src/main/java/io/cloudevents/core/builder/CloudEventBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    static CloudEventBuilder fromSpecVersion(SpecVersion version) {
+        switch (version) {
+            case V1:
+                return CloudEventBuilder.v1();
+            case V03:
+                return CloudEventBuilder.v03();
+        }
+        return null;
```
bsideup, line 51: Consider adding `@Nullable`

core/src/main/java/io/cloudevents/core/impl/CloudEventUtils.java
```diff
@@ -15,25 +15,28 @@
...
+     * for CloudEvent that doesn't implement CloudEventVisitable
+     *
+     * @param event the event to convert
+     * @return the visitable implementation
      */
-    <T> T visit(StructuredMessageVisitor<T> visitor) throws MessageVisitException, IllegalStateException;
-
-    default CloudEvent toEvent() throws MessageVisitException, IllegalStateException {
-        return this.visit(EventFormat::deserialize);
+    public static CloudEventVisitable toVisitable(CloudEvent event) {
```
bsideup, line 34: 👍

core/src/main/java/io/cloudevents/core/builder/CloudEventBuilder.java
slinkydeveloper: But that's a condition that can never happen, `SpecVersion` or is `V1` or it's `V3`. `SpecVersion version` on the other hand should be non null

slinkydeveloper: we should review all these annotation usages in all modules after this pr, I'm quite sure there are some places where we miss these

slinkydeveloper:
@bsideup & @olegz can you review this?

core/src/main/java/io/cloudevents/core/builder/CloudEventBuilder.java
bsideup: Maybe `throw IllegallStateException` then

slinkydeveloper: The only way this could happen is that there is misalignment between core and api versions, let me explicit it

slinkydeveloper: https://github.com/cloudevents/sdk-java/pull/157/commits/e3e08cc4e5e2361ce65c0ead45809ceb123daefb

duglin:
@bsideup thanks for the quick review!

olegz:
I'l second Sergey. The `api` module is clean and this PR reinforces clear its separation with `core`.
Approved

markfisher:
great to see this, thanks @slinkydeveloper!

PR #158, opened 2020-05-29
Reworked a bit the Extension interface

slinkydeveloper:
Look at #122

This change:

* Change the signature of `readFromEvent` to accept `CloudEventExtensions`
* Remove `asMap()` and let `Extension` extend `CloudEventExtensions`. Because the number of attributes for a materialized extension is always the same, allocating a map to provide an unstructrured view (needed to write the materialized extension to the cloud event builder) is useless. This solution makes more efficient the implementation of such materialized extensions

### DIFF
--- a/api/src/main/java/io/cloudevents/Extension.java
+++ b/api/src/main/java/io/cloudevents/Extension.java
@@ -40,7 +40,7 @@ public interface Extension {
      *
      * @param key the name of the extension attribute
      * @return the extension value
-     * @throws IllegalArgumentException if the key is not a valid attribute key for this extension
+     * @throws IllegalArgumentException if the key is unknown to this extension
      */
     @Nullable
     Object getValue(String key) throws IllegalArgumentException;
--- a/api/src/main/java/io/cloudevents/Extension.java
+++ b/api/src/main/java/io/cloudevents/Extension.java
@@ -27,6 +27,6 @@ public interface Extension extends CloudEventExtensions {
      *
      * @param extensions
      */
-    void readFromEvent(CloudEventExtensions extensions);
+    void readFrom(CloudEventExtensions extensions);
 
 }
--- a/api/src/main/java/io/cloudevents/Extension.java
+++ b/api/src/main/java/io/cloudevents/Extension.java
@@ -17,10 +17,16 @@
 
 package io.cloudevents;
 
+import io.cloudevents.lang.Nullable;
+
+import javax.annotation.ParametersAreNonnullByDefault;
+import java.util.Set;
+
 /**
  * Materialized CloudEvent Extension interface to read/write the extension attributes key/values.
  */
-public interface Extension extends CloudEventExtensions {
+@ParametersAreNonnullByDefault
+public interface Extension {
 
     /**
      * Fill this materialized extension with values from a {@link CloudEventExtensions} implementation
@@ -28,5 +34,21 @@ public interface Extension extends CloudEventExtensions {
      * @param extensions
      */
     void readFrom(CloudEventExtensions extensions);
+
+    /**
+     * Get the attribute of extension named {@code key}
+     *
+     * @param key the name of the extension attribute
+     * @return the extension value
+     */
+    @Nullable
+    Object getValue(String key);
+
+    /**
+     * Get the set of possible extension attribute keys
+     *
+     * @return set of possible extension attribute keys
+     */
+    Set<String> getKeys();
 
 }
--- a/api/src/main/java/io/cloudevents/Extension.java
+++ b/api/src/main/java/io/cloudevents/Extension.java
@@ -40,9 +40,10 @@ public interface Extension {
      *
      * @param key the name of the extension attribute
      * @return the extension value
+     * @throws IllegalArgumentException if the key is not a valid attribute key for this extension
      */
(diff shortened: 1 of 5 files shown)


slinkydeveloper:
@bsideup mind giving a look?

api/src/main/java/io/cloudevents/Extension.java
```diff
@@ -17,12 +17,16 @@
...
+ */
+public interface Extension extends CloudEventExtensions {
 
-    Map<String, Object> asMap();
+    /**
+     * Fill this materialized extension with values from a {@link CloudEventExtensions} implementation
+     *
+     * @param extensions
+     */
+    void readFromEvent(CloudEventExtensions extensions);
```
bsideup, line 30:
```suggestion
    void readFrom(CloudEventExtensions extensions);
```

api/src/main/java/io/cloudevents/Extension.java
```diff
@@ -17,12 +17,16 @@
...
 
-import java.util.Map;
-
-public interface Extension {
-
-    void readFromEvent(CloudEvent event);
+/**
+ * Materialized CloudEvent Extension interface to read/write the extension attributes key/values.
+ */
+public interface Extension extends CloudEventExtensions {
```
bsideup, line 23: This sounds wrong. `Extension extends CloudEventExtensions` smells.

Especially given that it will look like:
```java
Extension ext = ...;

ext.getExtension("OMG WTF!");
ext.getExtensionNames(); // whaaaat?
```

slinkydeveloper: how about:

```
public interface Extension {

    void readFrom(CloudEventExtensions extensions);

   Object getValue(String key);

   Set<String> getKeys();

}
```

bsideup: Much better :)

WDYT about `Iterable<String> getKeys();`? Since extension may define a dynamic set of keys, `Set` here can be a bit restrictive, and it does not look like we need to have `Set`'s characteristics in this case

slinkydeveloper: To me it doesn't sound explicative `Iterable<String>`, because `Iterable` can contain duplicates. Because we're returning a key set, to me it makes sense using the `Set<String>` type, exactly like i expect a Map to return the keys as a set

slinkydeveloper: On the other hand, I don't think it would be a bad idea to let this class extend `Iterable<Map.Entry<String, Object>>` or have a method like `iter()` that returns such `Iterable`

slinkydeveloper: And btw such `Iterable` can be default implemented

bsideup: `Iterable<Map.Entry<String, Object>>` would require allocations.

Okay, let's keep `Set` then.

slinkydeveloper: > Iterable<Map.Entry<String, Object>> would require allocations.

Sure, it's like "use it at your own risk" :smile:

core/src/main/java/io/cloudevents/core/extensions/DistributedTracingExtension.java
```diff
@@ -49,23 +51,31 @@ public void setTracestate(String tracestate) {
...
-        map.put(TRACESTATE, this.tracestate);
-        return Collections.unmodifiableMap(map);
+    public Object getValue(String extensionName) {
+        switch (extensionName) {
+            case TRACEPARENT:
+                return this.traceparent;
+            case TRACESTATE:
+                return this.tracestate;
+        }
+        return null;
```
bsideup, line 73: Just thinking out loud here:
Perhaps we should be more defensive and throw an error on unknown key. Should help debugging issues when wrong key is used, for example

slinkydeveloper: In that case, should we bring it to the interface? aka:

```
Object getValue(String key) throws IllegalArgumentException;
```

bsideup: Won't hurt, I suppose 👍

slinkydeveloper: like https://github.com/cloudevents/sdk-java/pull/158/commits/e2de52c46989b47aefa7c33f2b5ae91d733add57

api/src/main/java/io/cloudevents/Extension.java
```diff
@@ -17,12 +17,39 @@
...
+     * @param extensions
+     */
+    void readFrom(CloudEventExtensions extensions);
+
+    /**
+     * Get the attribute of extension named {@code key}
+     *
+     * @param key the name of the extension attribute
+     * @return the extension value
+     * @throws IllegalArgumentException if the key is not a valid attribute key for this extension
```
bsideup, line 43: How about this:
```suggestion
     * @throws IllegalArgumentException if the key is unknown to this extension
```

to avoid additional questions on "validity" and "attribute"

PR #167, opened 2020-06-04
Rename Visitor/Visitable to Writer/Reader

slinkydeveloper:
This PR revisit the names `CloudEventVisitor` and `CloudEventVisitable` to `CloudEventWriter` and `CloudEventReader`. Even if thes two interfaces represents a visitor pattern, they don't give the idea of their role inside the sdk.

`Reader` and `Writer` gives more meaning: you understand that these interfaes are meant to read something as event and write something as event.

This vastly improves the usage of the "low level" message APIs (example with vertx but it's the very same with kafka and other integrations):

```java
Message reader = VertxMessageReaderFactory.create(httpReq)
CloudEvent event = reader.toEvent()
While on the write side:
VertxHttpServerResponseMessageWriter.create(res)
  .writeStructured(event, jsonFormat)
```

```java
VertxHttpServerResponseMessageWriter.create(res)
  .writeBinary(event)
```

And direct transcoding:

```java
Message reader = VertxMessageReaderFactory.create(httpReq)
R written = reader.write(VertxHttpServerResponseMessageWriter.create(res))
```

Please note that most of this PR is about renaming, the crucial part is `io.cloudevents.visitor`  package (inside cloudevents-api) and `io.cloudevents.message` (inside cloudevents-core)


slinkydeveloper:
My proposal is to name the package `io.cloudevents.rw`

Issue #175, opened 2020-06-17
Sub-second resolution for `time` attribute.


matejvasek:
Deserialization  crashes when message contains time with fractional seconds, e.g. 2018-04-26T14:48:09.1234Z.

PR #176, opened 2020-06-17
fix: sub-second resolution `time` attribute

matejvasek:
Fixes parsing of dates like: 2018-04-26T14:48:09.1234Z

Fixes: #175


slinkydeveloper:
@matejvasek thanks for your contribution! Can we have a test in the api package too to check correct `parseTime()` behaviour?

PR #181, opened 2020-06-18
Javadoc-ed cloudevents-api

slinkydeveloper:
Did another round of Javadoc to `cloudevents-api` module.

Also changed and clarified the `getAttributeNames()` semantic (the previous semantic was useless, because it was equal to `getSpecVersion().getAllAttributes()`)

### DIFF
--- a/api/src/main/java/io/cloudevents/CloudEventAttributes.java
+++ b/api/src/main/java/io/cloudevents/CloudEventAttributes.java
@@ -23,6 +23,7 @@ import javax.annotation.ParametersAreNonnullByDefault;
 import java.net.URI;
 import java.time.ZonedDateTime;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * Interface which defines CloudEvent attributes as per specification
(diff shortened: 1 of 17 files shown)


api/src/main/java/io/cloudevents/CloudEventAttributes.java
```diff
@@ -78,9 +79,25 @@
...
+     */
     @Nullable
-    Object getAttribute(String extensionName);
+    Object getAttribute(String attributeName) throws IllegalArgumentException;
 
-    Set<String> getAttributeNames();
+    /**
+     * @return The non-null <a href="https://github.com/cloudevents/spec/blob/v1.0/spec.md#context-attributes">context attributes</a> names in this instance
+     */
+    default Set<String> getAttributeNames() {
```
bsideup, line 95: do we even need this method then? I'd propose to drop it , since `getSpecVersion().getAllAttributes()` can be used instead

core/src/main/java/io/cloudevents/core/v03/CloudEventV03.java
```diff
@@ -99,8 +98,10 @@ public ZonedDateTime getTime() {
     }
 
     @Override
-    public Object getAttribute(String name) {
-        switch (name) {
+    public Object getAttribute(String attributeName) {
+        switch (attributeName) {
+            case "specversion":
+                return SpecVersion.V03;
```
bsideup, line 104:
```suggestion
                return getSpecVersion();
```

core/src/main/java/io/cloudevents/core/v1/CloudEventV1.java
```diff
@@ -93,8 +92,10 @@ public ZonedDateTime getTime() {
     }
 
     @Override
-    public Object getAttribute(String name) {
-        switch (name) {
+    public Object getAttribute(String attributeName) {
+        switch (attributeName) {
+            case "specversion":
+                return SpecVersion.V1;
```
bsideup, line 98:
```suggestion
                return getSpecVersion();
```

api/src/main/java/io/cloudevents/CloudEventAttributes.java
slinkydeveloper: I have no strong opinion on this, i think it can be useful for a user that wants to get only the attributes available in this instance, and it doesn't cost a lot to maintain it

PR #208, opened 2020-08-11
[WIP] Allow POJOs as CloudEvent data, round 2

### LINKED ISSUE #196: How to best deal with conversions/seralization to other formats that byte[]?
I'm experimenting with using cloudevents as the storage format for an [event sourcing](https://martinfowler.com/eaaDev/EventSourcing.html) library I'm building (you can find the project [here](https://github.com/johanhaleby/occurrent)). However when I upgraded to `io.cloudevents:cloudevents-json-jackson:2.0.0-milestone1` from `1.3.0` I noticed that the "data" field is no longer generic but is now always `byte[]`. I.e. where you previously could do:

```java
builder.withData(Map.of("hello", "World")). ..
```

you now need to convert the map into a `byte[]`:

```java
builder.withData(objectMapper.writeValueAsBytes(Map.of("hello", "World"))). ..
```

Also `EventFormat` has changed to only deal with `byte[]`. 

I actually think I like this change but it comes with drawbacks in some cases. In my case I'm trying to serialize the CloudEvents to be stored in MongoDB. Since MongoDB understands json natively I'd like to convert the `CloudEvent` instance to a model that MongoDB Java Driver understands, for example a [org.bson.Document](https://mongodb.github.io/mongo-java-driver/4.0/apidocs/bson/org/bson/Document.html). Currently, if I want to use the `io.cloudevents.jackson.JsonFormat`, I need to do double encoding:

```java
CloudEvent ce = ..
byte[] serializedCloudEventAsByteArray = jsonEventFormat.serialize(ce);
// I want to store a Document so I need to convert again
String serializedCloudEventAsString = new String(bytes, UTF_8);
Document cloudEventAsDocument = Document.parse(serializedCloudEventAsString);
mongo.insert(cloudEventAsDocument);
```

This involves several unnecessary steps. Since the [Document.parse](https://mongodb.github.io/mongo-java-driver/4.0/apidocs/bson/org/bson/Document.html#parse(java.lang.String)) method only takes  a `String` I need to first create a `String` from the `byte[]` produced by the `EventFormat` and then pass it to `Document.parse` which will _parse_ the JSON string and convert it into a `Document`.

Compare that with using jackson directly where you could simply do:

```java
CloudEvent ce = ..
Document document = objectMapper.convertValue(ce, Document.class);
mongo.insert(cloudEventAsDocument);
```

which is much more efficient but prohibited when using an instance of `EventFormat`. 

**Questions**

How should one think when dealing with issues as the one above? I think the cloud event spec fit well as storage format but the current implementation prohibits efficient mapping to datastore model (`org.bson.Document` in this case). Should you avoid using `EventFormat` in these cases and write custom serialization/deserialization code?

For reference I've actually copied `io.cloudevents.jackson.JsonFormat` into a custom `DocumentFormat` (which cannot implement `EventFormat` since the types match) to demonstrate the point. As you can see I cannot implement `EventFormat` since it requires a byte array. Other than that it's virtually identical to the existing `JsonFormat` but instead of doing `objectMapper.writeValueAsBytes(..)` I use the `objectMapper.convertValue(..)` method.

```java
package se.haleby.occurrent.eventstore.mongodb.converter;

import com.fasterxml.jackson.databind.ObjectMapper;
import com.fasterxml.jackson.databind.module.SimpleModule;
import io.cloudevents.CloudEvent;
import io.cloudevents.core.format.EventDeserializationException;
import io.cloudevents.core.format.EventFormat;
import io.cloudevents.core.format.EventSerializationException;
import io.cloudevents.jackson.CloudEventDeserializer;
import io.cloudevents.jackson.CloudEventSerializer;
import org.bson.Document;

public final class DocumentFormat {

    public static final String CONTENT_TYPE = "application/cloudevents+json";

    private final ObjectMapper mapper;
    private final boolean forceDataBase64Serialization;
    private final boolean forceStringSerialization;

    public DocumentFormat(boolean forceDataBase64Serialization, boolean forceStringSerialization) {
        this.mapper = new ObjectMapper();
        this.mapper.registerModule(getCloudEventJacksonModule(forceDataBase64Serialization, forceStringSerialization));
        this.forceDataBase64Serialization = forceDataBase64Serialization;
        this.forceStringSerialization = forceStringSerialization;
    }

    public DocumentFormat() {
        this(false, false);
    }

    /**
     * @return a copy of this JsonFormat that serialize events with json data with Base64 encoding
     */
    public DocumentFormat withForceJsonDataToBase64() {
        return new DocumentFormat(true, this.forceStringSerialization);
    }

    /**
     * @return a copy of this JsonFormat that serialize events with non-json data as string
     */
    public DocumentFormat withForceNonJsonDataToString() {
        return new DocumentFormat(this.forceDataBase64Serialization, true);
    }

    public Document serialize(CloudEvent event) throws EventSerializationException {
        return mapper.convertValue(event, Document.class);
    }

    public CloudEvent deserialize(Document document) throws EventDeserializationException {
        return mapper.convertValue(document, CloudEvent.class);
    }

    public String serializedContentType() {
        return CONTENT_TYPE;
    }

    /**
     * @return a JacksonModule with CloudEvent serializer/deserializer with default values
     */
    public static SimpleModule getCloudEventJacksonModule() {
        return getCloudEventJacksonModule(false, false);
    }

    /**
     * @return a JacksonModule with CloudEvent serializer/deserializer customizing the data serialization.
     * Look at {@link #withForceJsonDataToBase64()} and {@link #withForceNonJsonDataToString()} for more details.
     */
    public static SimpleModule getCloudEventJacksonModule(boolean forceDataBase64Serialization, boolean forceStringSerialization) {
        final SimpleModule ceModule = new SimpleModule("CloudEvent");
        ceModule.addSerializer(CloudEvent.class, new CloudEventSerializer(forceDataBase64Serialization, forceStringSerialization) {
        });
        ceModule.addDeserializer(CloudEvent.class, new CloudEventDeserializer() {
        });
        return ceModule;
    }

    protected static boolean dataIsJsonContentType(String contentType) {
        // If content type, spec states that we should assume is json
        return contentType == null || contentType.startsWith("application/json") || contentType.startsWith("text/json");
    }
}

slinkydeveloper:
Fix #196 and follow up from discussions in #198

In this second iteration I've created a new `CloudEvent` interface in core that extends the one in `api`. This interface adds `getRawData()` This interface is used mostly everywhere in core and that's what the user of the sdk should use too (leaving the one in api mostly for integration needs), but always paying attention to support the parent `CloudEvent` when possible.

As in #198, I had to change the reader/writer interface and I added `EventDataCodec` but only with serialization support, deserialization of data is out of the scope of this PR (and of this project too)


olegz:
The #198 has suggested a couple of approaches of adding additional functionality to core strategies - approaches that are intuitive and widely used in many java frameworks. So I am not sure why do keep on insisting doing something completely different. . .??

At the very least, **please, please, please**, don't call it `CloudEvent`. Aside form it being an anti-pattern, it will create a tremendous confusion and a guaranteed CCEs when developers declare `Something<CloudEvent>`.

slinkydeveloper:
> The #198 has suggested a couple of approaches of adding additional functionality to core strategies - approaches that are intuitive and widely used in many java frameworks. So I am not sure why do keep on insisting doing something completely different. . .??

You're free to contribute with a PR, in order to propose a solution that you think might fit the needs discussed in #196 and #198

olegz:
My comments were simply meant to address the issues I see with the current PR, but I certainly appreciate the invitation to collaborate. . .

As suggested in #198, by [several people](https://github.com/cloudevents/sdk-java/pull/198#issuecomment-670505615), all you need to do is define additional method(s) on `BaseCloudEvent` (e.g., `BaseCloudEvent.getRawData()`) or some other specialized implementation within core module. 

Basically remove `CloudEvent` interface from the core and add `getRawData()` method on `BaseCloudEvent` implementation. 
OR
Rename `CloudEvent` in core to something else to avoid name-space/type collision issues.

In it's current state my vote (if counts) is -1

slinkydeveloper:
I think we still need a new interface for that if a 3rd party implementer wants to use the features of core...

bsideup:
> and actually no name is better than `CloudEvent`

- `TypedCloudEvent`?
- `CarrierCloudEvent`?
- `POJOCloudEvent`?

I am sure there are 10s of other suitable names, but naming it `CloudEvent` in `core` while there is one in `api` is a terrible idea.
Anyone who worked with Java long enough remembers confusing `List` and `List` or `Date` and `Date` (I let the reader to figure out the package names ;))

Since the need for this interface is questionable anyways, let's _at least_ not repeat the same mistake with clashing names.

slinkydeveloper:
None of the names you discuss really makes sense IMO, because the `CloudEvent` as entity is always the same and the fact that there is a new interface to explicit the idea of having the `data` field as `Object` it's unnatural: the payload of a cloudevent is dynamic by nature, to the cloudevent user (not the java user) `TypedCloudEvent` doesn't really makes sense, what is typed? the payload? how is typed?

Hence my proposal in #211, relax the `getData()` and we solve the issue once and for all

bsideup:
@slinkydeveloper 

> a new interface to explicit the idea of having the `data` field as `Object` it's unnatural: the payload of a cloudevent is dynamic by nature

You should not forget that you're dealing with Java, where certain restrictions apply that you have to deal with.

> `TypedCloudEvent` doesn't really makes sense, what is typed? the payload? how is typed?

If not the payload, what else can be typed? The whole `CloudEvent` thing is around the payload, with some extra attributes/extensions that help understanding the payload better. If you think about it this way, it is much easier to reason about the naming and what `TypedCloudEvent` represents.

olegz:
One naming pattern that could serve as viable compromise and the one that is widely used is `*Aware`. In Spring we use it in many places (i can share examples if needed). For example `TypeAwareCloudEvent` or `RawDataAwareCloudEvent`

slinkydeveloper:
Replaced by https://github.com/cloudevents/sdk-java/pull/250

PR #216, opened 2020-08-31
Switch `ZonedDateTime` to `OffsetDateTime`

### LINKED ISSUE #214: Switch `ZonedDateTime` to `OffsetDateTime`
The cloudevents spec clearly [says](https://github.com/cloudevents/spec/blob/v1.0/spec.md#time) that the `time` attribute "MUST adhere to the format specified in [RFC 3339](https://tools.ietf.org/html/rfc3339)".  The `java.time.xxx` class that corresponds to RFC 3339 is `OffsetDateTime`.

`ZonedDateTime` adds a time zone identifier (in the form of region/city) to the end of the string on serialization, meaning that an event containing a time output by normal serialization does not conform the the spec.

`OffsetDateTime` is easier to work with, takes less space and more correctly implements the spec.

slinkydeveloper:
Fix #214

There is also a small change that prevented travis to correctly build the module (while building, the cache was using an old version of cloudevents-api)

Mind giving a look @pwall-tyro ?

### DIFF
--- a/api/src/main/java/io/cloudevents/types/Time.java
+++ b/api/src/main/java/io/cloudevents/types/Time.java
@@ -18,10 +18,7 @@
 package io.cloudevents.types;
 
 import java.time.OffsetDateTime;
-import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.DateTimeParseException;
-import java.time.temporal.ChronoField;
 
 /**
  * Utilities to handle the <a href="https://github.com/cloudevents/spec/blob/v1.0/spec.md#type-system">CloudEvent Attribute Timestamp type</a>
@@ -31,23 +28,17 @@ public final class Time {
     private Time() {
     }
 
-    public static final DateTimeFormatter RFC3339_DATE_FORMAT = new DateTimeFormatterBuilder()
-        .appendPattern("yyyy-MM-dd'T'HH:mm:ss")
-        .appendFraction(ChronoField.NANO_OF_SECOND, 0, 9, true)
-        .appendOffsetId()
-        .toFormatter();
-
     /**
      * Parse a {@link String} RFC3339 compliant as {@link OffsetDateTime}
      */
     public static OffsetDateTime parseTime(String time) throws DateTimeParseException {
-        return OffsetDateTime.parse(time, RFC3339_DATE_FORMAT);
+        return OffsetDateTime.parse(time);
     }
 
     /**
      * Convert a {@link OffsetDateTime} to {@link String}
      */
     public static String writeTime(OffsetDateTime time) throws DateTimeParseException {
-        return time.format(RFC3339_DATE_FORMAT);
+        return time.toString();
     }
 }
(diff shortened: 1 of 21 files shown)


api/src/main/java/io/cloudevents/types/Time.java
```diff
@@ -38,16 +38,16 @@ private Time() {
         .toFormatter();
 
     /**
-     * Parse a {@link String} RFC3339 compliant as {@link ZonedDateTime}
+     * Parse a {@link String} RFC3339 compliant as {@link OffsetDateTime}
      */
-    public static ZonedDateTime parseTime(String time) throws DateTimeParseException {
-        return ZonedDateTime.parse(time, RFC3339_DATE_FORMAT);
+    public static OffsetDateTime parseTime(String time) throws DateTimeParseException {
+        return OffsetDateTime.parse(time, RFC3339_DATE_FORMAT);
```
pwall-tyro, line 44: The default formatter used by `OffsetDateTime` is `DateTimeFormatter.ISO_OFFSET_DATE_TIME`, which conforms exactly to the RFC3339 "date-time" specification.  It is therefore unnecessary to use a custom formatter;
```
    OffsetDateTime.parse(time);
```
will achieve the same result more simply and clearly.

Similarly,
```
    offsetDateTimeValue.toString();
```
will output `OffsetDateTime` values in RFC3339 form without the need to specify a formatter.  The Jackson Java 8 Date/time module also uses this format by default for `OffsetDateTime`.  (This is one of the reasons for recommending the change from `ZonedDateTime` to `OffsetDateTime`.)

pwall-tyro (approved):
Thank you for taking my suggestion on board.  I have made a recommendation that may simplify the code a little further, but I am happy to approve the changes in any case.

api/src/main/java/io/cloudevents/types/Time.java
slinkydeveloper: lemme try :smile:

slinkydeveloper:
@pwall-tyro can you check it again? :smile:

pwall-tyro (approved):
:+1:

api/src/test/java/io/cloudevents/types/TimeTest.java
```diff
@@ -46,18 +46,18 @@ void testParseDateOffset() {
...
     }
 
     public static Stream<Arguments> parseDateArguments() {
         return Stream.of(
-            Arguments.of("1985-04-12T23:20:50.52Z"),
-            Arguments.of("1990-12-31T23:59:00Z"),
-            Arguments.of("1990-12-31T15:59:00-08:00"),
-            Arguments.of("1937-01-01T12:00:27.87+00:20")
+            Arguments.of("1985-04-12T23:20:50.520Z"),
+            Arguments.of("1990-12-31T23:59Z"),
```
matejvasek, line 58: @slinkydeveloper is this a valid RFC3339 date if there are no seconds?

slinkydeveloper: I was looking at the abnf, it seems like i made a mistake here

```
   partial-time    = time-hour ":" time-minute ":" time-second
                     [time-secfrac]
   full-date       = date-fullyear "-" date-month "-" date-mday
   full-time       = partial-time time-offset
```

partial-time has time-second not optional

Issue #207, opened 2020-08-11
Consider replacing byte[] in getData() with ByteBuffer


slinkydeveloper:
`ByteBuffer`s gives a great deal of flexibility (they can be allocated in and off heap). If allocated outside the heap, they give a huge perf boost and, if allocated in the heap, they have more or less the same cost of a `byte[]`.

Replacing `byte[]` will require changing the `MessageReader` and `MessageWriter` interfaces too, together with all the impls

slinkydeveloper:
Should we do it?

slinkydeveloper:
The problem with `ByteBuffer` is that there is a whole complexity involved when the buffer is file mapped and you need to literally "consume it" and load in memory, in a `byte[]` or in another ByteBuffer memory mapped. To me this sounds too low level for our users. WDYT?

pierDipi:
With Kafka's [KIP-863: Reduce CompletedFetch#parseRecord() memory copy](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=225152035) this might be more relevant now

PR #225, opened 2020-09-01
Generic HTTP bindings

### LINKED ISSUE #212: Much harder with V2 than V1 to send binary CloudEvents
I might be missing something, but it seems to me that it has become much harder to send binary CloudEvents with the V2 SDK, unless you happen to be using one of the small number of supported HTTP clients. In V1, I could write this:

```
Wire<String, String, String> wire = Marshallers.<Map<String, Object>>binary()
        .withEvent(() -> myCloudEvent)
        .marshal();
```

Then I basically just needed to copy `wire.getHeaders()` into the HTTP headers of the request and have `wire.getPayload().get()` be the HTTP body. That's easy to do with pretty much any HTTP API.

With V2 it looks like I am out of luck if there isn't already a `MessageWriter` defined for the specific HTTP API that I am using. It's not _that_ hard to make a `MessageWriter` by following the example of one of the existing ones, but it's still about 50 lines of non-trivial code.

Could we have the simplicity of V1 back? Perhaps with a trivial MessageWriter implementation that just stashes the HTTP headers in a `Map<String, String>` and the body in a `byte[]` or whatever, which is more or less what the old `Wire` did. It's nice that the supported HTTP clients are so easy to use, but it seems to be at the cost of making anything else much harder.

matejvasek:
resolves #212

### DIFF
--- a/examples/basic-http/pom.xml
+++ b/examples/basic-http/pom.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2018-Present The CloudEvents Authors
+  ~ <p>
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~ <p>
+  ~ http://www.apache.org/licenses/LICENSE-2.0
+  ~ <p>
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>cloudevents-examples</artifactId>
+        <groupId>io.cloudevents</groupId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>cloudevents-basic-http-example</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-http-basic</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-json-jackson</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-server</artifactId>
+            <version>9.4.28.v20200408</version>
+        </dependency>
+    </dependencies>
+</project>
--- a/examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/IOUtils.java
+++ b/examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/IOUtils.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2018-Present The CloudEvents Authors
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ package io.cloudevents.examples.http.basic;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+// based on apache's commons-io
+// which is licensed under the Apache License, Version 2.0
+// https://commons.apache.org/proper/commons-io/
+// https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/IOUtils.java
+
+public final class IOUtils {
+
+    private IOUtils() {}
+
+    // since Java 9+ you may call InputStream.readAllBytes() instead of this method
+    public static byte[] toByteArray(InputStream body) throws IOException {
+        try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
+            byte[] buff = new byte[(1<<10) * 8];
+            int read;
+            while ((read = body.read(buff)) != -1) {
+                byteArrayOutputStream.write(buff, 0, read);
+            }
+            return byteArrayOutputStream.toByteArray();
+        }
+    }
+}
--- a/http/basic/README.md
+++ b/http/basic/README.md
@@ -1,0 +1,42 @@
+# Generic HTTP Protocol Binding
+
+[![Javadocs](http://www.javadoc.io/badge/io.cloudevents/cloudevents-http-basic.svg?color=green)](http://www.javadoc.io/doc/io.cloudevents/cloudevents-http-basic)
+
+This module is designed to be usable with various HTTP APIs.
+
+There are also more specialized HTTP bindings:
+* [`cloudevents-http-vertx`](../vertx)
+* [`cloudevents-http-restful-ws`](../restful-ws)
+
+Since this module is generic it doesn't offer optimal performance for all HTTP implementations.
+For better performance consider implementing `MessageReader` and `MessageWriter` that are
+tailored for specific HTTP implementation. As a reference you can take aforementioned existing bindings.
+
+For Maven based projects, use the following to configure the CloudEvents Generic HTTP Transport:
+
+```xml
+<dependency>
+    <groupId>io.cloudevents</groupId>
+    <artifactId>cloudevents-http-basic</artifactId>
+    <version>2.0.0-milestone1</version>
+</dependency>
+```
+
+## Sending and Receiving CloudEvents
+
+To send and receive CloudEvents we use `MessageWriter` and `MessageReader`, respectively.
+This module offers factory methods for creation of those in `HttpMessageFactory`.
+
+```java
+public class HttpMessageFactory {
+    public static MessageReader createReader(Consumer<BiConsumer<String,String>> forEachHeader, byte[] body);
+    public static MessageReader createReader(Map<String,String> headers, byte[] body);
+    public static MessageReader createReaderFromMultimap(Map<String,List<String>> headers, byte[] body);
+    public static MessageWriter createWriter(BiConsumer<String, String> putHeader, Consumer<byte[]> sendBody);
+}
+```
+
+Examples of usage:
+* [Standard Java HttpServer](../../examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/HttpServer.java)
+* [Http Client with HttpURLConnection](../../examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/HttpURLConnectionClient.java)
+* [Http Servlet with Jetty](../../examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/JettyServer.java)
--- a/http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
+++ b/http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2018-Present The CloudEvents Authors
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cloudevents.http;
+
+import io.cloudevents.core.message.MessageReader;
+import io.cloudevents.core.message.MessageWriter;
+import io.cloudevents.core.message.impl.GenericStructuredMessageReader;
+import io.cloudevents.core.message.impl.MessageUtils;
+import io.cloudevents.core.message.impl.UnknownEncodingMessageReader;
+import io.cloudevents.http.impl.CloudEventsHeaders;
+import io.cloudevents.http.impl.HttpMessageReader;
+import io.cloudevents.http.impl.HttpMessageWriter;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+
+import static io.cloudevents.http.impl.CloudEventsHeaders.CONTENT_TYPE;
+
+/**
+ * This class provides a collection of methods to create {@link io.cloudevents.core.message.MessageReader}
+ * and {@link io.cloudevents.core.message.MessageWriter} for various HTTP APIs.
+ */
+public final class HttpMessageFactory {
+
+    private HttpMessageFactory() {}
+
+    /**
+     * Creates a new {@link MessageReader} that can read both structured and binary messages from a HTTP response (client) or request (server).
+     *
+     * <pre>
+     * Example of usage with <a href="https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletRequest.html">HttpServletRequest</a>:
+     * {@code
+     * Consumer<BiConsumer<String,String>> forEachHeader = processHeader -> {
+     *     Enumeration<String> headerNames = httpServletRequest.getHeaderNames();
+     *     while (headerNames.hasMoreElements()) {
+     *         String name = headerNames.nextElement();
+     *         processHeader.accept(name, httpServletRequest.getHeader(name));
+     *
+     *     }
+     * };
+     * byte[] body = httpServletRequest.getInputStream().readAllBytes();
+     * HttpMessageFactory.createReader(forEachHeader, body);
+     * }
+     * </pre>
+     * @param forEachHeader http headers visitor function
+     * @param body          nullable buffer of the body
+     * @return a message reader implementation with potentially an unknown encoding
+     * @throws IllegalArgumentException If, in case of binary mode, the spec version is invalid
+     */
+    public static MessageReader createReader(Consumer<BiConsumer<String, String>> forEachHeader, byte[] body) {
+        final AtomicReference<String> contentType = new AtomicReference<>();
+        final AtomicReference<String> specVersion = new AtomicReference<>();
+
+        forEachHeader.accept((k, v) -> {
+            if (CONTENT_TYPE.equalsIgnoreCase(k)) {
+                contentType.set(v);
+            } else if (CloudEventsHeaders.SPEC_VERSION.equalsIgnoreCase(k)) {
+                specVersion.set(v);
+            }
+        });
+
+        return MessageUtils.parseStructuredOrBinaryMessage(
+            contentType::get,
+            format -> new GenericStructuredMessageReader(format, body),
+            specVersion::get,
+            sv -> new HttpMessageReader(sv, forEachHeader, body),
+            UnknownEncodingMessageReader::new
+        );
+    }
+
+    /**
+     * Creates a new {@link MessageReader} that can read both structured and binary messages from a HTTP response (client) or request (server).
+     *
+     * <pre>
+     * This overload is equivalent to calling:
+     * {@code HttpMessageFactory.createReader(headers::forEach, body);}
+     * </pre>
+     * @param headers http headers as map
+     * @param body    nullable buffer of the body
+     * @return a message reader implementation with potentially an unknown encoding
+     * @throws IllegalArgumentException If, in case of binary mode, the spec version is invalid
+     */
+    public static MessageReader createReader(Map<String, String> headers, byte[] body) {
+        return createReader(headers::forEach, body);
+    }
+
+    /**
+     * Creates a new {@link MessageReader} that can read both structured and binary messages from a HTTP response (client) or request (server).
+     *
+     * @param headers http headers as multimap
+     * @param body    nullable buffer of the body
+     * @return a message reader implementation with potentially an unknown encoding
+     * @throws IllegalArgumentException If, in case of binary mode, the spec version is invalid
+     */
+    public static MessageReader createReaderFromMultimap(Map<String, List<String>> headers, byte[] body) {
+        Consumer<BiConsumer<String, String>> forEachHeader = processHeader ->
+            headers.forEach((key, values) ->
+                values.forEach(value -> processHeader.accept(key, value)));
+        return createReader(forEachHeader, body);
+    }
+
+    /**
+     * Creates a new {@link MessageWriter} that can write both structured and binary messages to a HTTP response (server) or request (client).
+     *
+     * <pre>
+     * Example of usage with <a href="https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletResponse.html">HttpServletResponse</a>:
+     * {@code
+     * HttpMessageFactory.createWriter(
+     *     httpServletResponse::addHeader,
+     *     body -> {
+     *         try {
+     *             if (body != null) {
+     *                 httpServletResponse.setContentLength(body.length);
+     *                 httpServletResponse.setStatus(HttpStatus.OK_200);
+     *                 try (ServletOutputStream outputStream = httpServletResponse.getOutputStream()) {
+     *                     outputStream.write(body);
+     *                 }
+     *             } else {
+     *                 httpServletResponse.setStatus(HttpStatus.NO_CONTENT_204);
+     *             }
+     *         } catch (IOException e) {
+     *             throw new UncheckedIOException(e);
+     *         }
+     * });
+     * }
+     * </pre>
+     * @param putHeader a function that puts header into HTTP request or response.
+     * @param sendBody  a function that sends body (e.g. sets HTTP status code, content-length and writes the bytes into output stream).
+     * @return a message writer
+     */
+    public static MessageWriter createWriter(BiConsumer<String, String> putHeader, Consumer<byte[]> sendBody) {
+        return new HttpMessageWriter(putHeader, sendBody);
+    }
+
+}
--- a/http/basic/src/main/java/io/cloudevents/http/impl/CloudEventsHeaders.java
+++ b/http/basic/src/main/java/io/cloudevents/http/impl/CloudEventsHeaders.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2018-Present The CloudEvents Authors
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cloudevents.http.impl;
+
+import io.cloudevents.core.message.impl.MessageUtils;
+
+import java.util.Collections;
+import java.util.Map;
+
+public final class CloudEventsHeaders {
+
+    private CloudEventsHeaders() {}
+
+    public static final String CONTENT_TYPE = "Content-Type";
+
+    public static final String CE_PREFIX = "ce-";
+
+    public static final Map<String, String> ATTRIBUTES_TO_HEADERS = Collections.unmodifiableMap(MessageUtils.generateAttributesToHeadersMapping(v -> {
+        if (v.equals("datacontenttype")) {
+            return CONTENT_TYPE;
+        }
+        return CE_PREFIX + v;
+    }));
+
+    public static final String SPEC_VERSION = ATTRIBUTES_TO_HEADERS.get("specversion");
+
+}
--- a/http/basic/src/test/java/io/cloudevents/http/HttpMessageReaderWriterTest.java
+++ b/http/basic/src/test/java/io/cloudevents/http/HttpMessageReaderWriterTest.java
@@ -1,0 +1,248 @@
+/*
+ * Copyright 2018-Present The CloudEvents Authors
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cloudevents.http;
+
+import io.cloudevents.CloudEvent;
+import io.cloudevents.SpecVersion;
+import io.cloudevents.core.message.Encoding;
+import io.cloudevents.core.message.MessageReader;
+import io.cloudevents.core.message.impl.GenericStructuredMessageReader;
+import io.cloudevents.core.mock.CSVFormat;
+import io.cloudevents.http.HttpMessageFactory;
+import io.cloudevents.http.impl.HttpMessageWriter;
+import io.cloudevents.types.Time;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Stream;
+
+import static io.cloudevents.core.test.Data.*;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class HttpMessageReaderWriterTest {
+
+    @ParameterizedTest
+    @MethodSource("io.cloudevents.core.test.Data#allEventsWithoutExtensions")
+    void testWriteStructured(CloudEvent event) {
+
+        final AtomicReference<byte[]> body = new AtomicReference<>();
+        final Map<String, String> headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+        GenericStructuredMessageReader.from(event, CSVFormat.INSTANCE).visit(
+            HttpMessageFactory.createWriter(headers::put, body::set)
+        );
+        assertThat(headers.get("content-type"))
+            .isEqualTo(CSVFormat.INSTANCE.serializedContentType());
+        assertThat(body.get())
+            .isEqualTo(CSVFormat.INSTANCE.serialize(event));
+
+    }
+
+    @ParameterizedTest
+    @MethodSource("binaryTestArguments")
+    void testWriteBinary(CloudEvent event, Map<String,String> expectedHeaders, byte[] expectedBody) {
+
+        final AtomicReference<byte[]> body = new AtomicReference<>();
+        final Map<String, String> headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+
+        new HttpMessageWriter(headers::put, body::set).writeBinary(event);
+
+        expectedHeaders.forEach((k, v) -> {
+            assertThat(headers.get(k)).isEqualTo(v);
+        });
+        if (expectedBody != null) {
+            assertThat(body.get()).isEqualTo(expectedBody);
+        }
+    }
+
+    @ParameterizedTest
+    @MethodSource("io.cloudevents.core.test.Data#allEventsWithoutExtensions")
+    public void testReadStructured(CloudEvent event) {
+        byte[] serializedEvent = CSVFormat.INSTANCE.serialize(event);
+
+        Map<String,String> headers = new HashMap<String,String>() {{
+            put("content-type", CSVFormat.INSTANCE.serializedContentType() + "; charset=utf8");
+        }};
+
+        MessageReader message = HttpMessageFactory.createReader(
+            headers,
+            serializedEvent
+        );
+
+        assertThat(message.getEncoding())
+            .isEqualTo(Encoding.STRUCTURED);
+        assertThat(message.toEvent())
+            .isEqualTo(event);
+    }
+
+    @ParameterizedTest
+    @MethodSource("binaryTestArguments")
+    public void testReadBinary(CloudEvent expectedEvent, Map<String,String> headers, byte[] body) {
+        MessageReader message = HttpMessageFactory.createReader(headers::forEach, body);
+
+        assertThat(message.getEncoding())
+            .isEqualTo(Encoding.BINARY);
+        assertThat(message.toEvent())
+            .isEqualTo(expectedEvent);
+    }
+
+    public static Stream<Arguments> binaryTestArguments() {
+        return Stream.of(
+            // V03
+            Arguments.of(
+                V03_MIN,
+                createHeaders()
+                    .add("ce-specversion", SpecVersion.V03.toString())
+                    .add("ce-id", ID)
+                    .add("ce-type", TYPE)
+                    .add("ce-source", SOURCE.toString()),
+                null
+            ),
+            Arguments.of(
+                V03_WITH_JSON_DATA,
+                createHeaders()
+                    .add("ce-specversion", SpecVersion.V03.toString())
+                    .add("ce-id", ID)
+                    .add("ce-type", TYPE)
+                    .add("ce-source", SOURCE.toString())
+                    .add("ce-schemaurl", DATASCHEMA.toString())
+                    .add("content-type", DATACONTENTTYPE_JSON)
+                    .add("ce-subject", SUBJECT)
+                    .add("ce-time", Time.writeTime(TIME)),
+                DATA_JSON_SERIALIZED
+            ),
+            Arguments.of(
+                V03_WITH_JSON_DATA_WITH_EXT_STRING,
+                createHeaders()
+                    .add("ce-specversion", SpecVersion.V03.toString())
+                    .add("ce-id", ID)
+                    .add("ce-type", TYPE)
+                    .add("ce-source", SOURCE.toString())
+                    .add("ce-schemaurl", DATASCHEMA.toString())
+                    .add("content-type", DATACONTENTTYPE_JSON)
+                    .add("ce-subject", SUBJECT)
+                    .add("ce-time", Time.writeTime(TIME))
+                    .add("ce-astring", "aaa")
+                    .add("ce-aboolean", "true")
+                    .add("ce-anumber", "10"),
+                DATA_JSON_SERIALIZED
+            ),
+            Arguments.of(
+                V03_WITH_XML_DATA,
+                createHeaders()
+                    .add("ce-specversion", SpecVersion.V03.toString())
+                    .add("ce-id", ID)
+                    .add("ce-type", TYPE)
+                    .add("ce-source", SOURCE.toString())
+                    .add("content-type", DATACONTENTTYPE_XML)
+                    .add("ce-subject", SUBJECT)
+                    .add("ce-time", Time.writeTime(TIME)),
+                DATA_XML_SERIALIZED
+            ),
+            Arguments.of(
+                V03_WITH_TEXT_DATA,
+                createHeaders()
+                    .add("ce-specversion", SpecVersion.V03.toString())
+                    .add("ce-id", ID)
+                    .add("ce-type", TYPE)
+                    .add("ce-source", SOURCE.toString())
+                    .add("content-type", DATACONTENTTYPE_TEXT)
+                    .add("ce-subject", SUBJECT)
+                    .add("ce-time", Time.writeTime(TIME)),
+                DATA_TEXT_SERIALIZED
+            ),
+            // V1
+            Arguments.of(
+                V1_MIN,
+                createHeaders()
+                    .add("ce-specversion", SpecVersion.V1.toString())
+                    .add("ce-id", ID)
+                    .add("ce-type", TYPE)
+                    .add("ce-source", SOURCE.toString()),
+                null
+            ),
+            Arguments.of(
+                V1_WITH_JSON_DATA,
+                createHeaders()
+                    .add("ce-specversion", SpecVersion.V1.toString())
+                    .add("ce-id", ID)
+                    .add("ce-type", TYPE)
+                    .add("ce-source", SOURCE.toString())
+                    .add("ce-dataschema", DATASCHEMA.toString())
+                    .add("content-type", DATACONTENTTYPE_JSON)
+                    .add("ce-subject", SUBJECT)
+                    .add("ce-time", Time.writeTime(TIME)),
+                DATA_JSON_SERIALIZED
+            ),
+            Arguments.of(
+                V1_WITH_JSON_DATA_WITH_EXT_STRING,
+                createHeaders()
+                    .add("ce-specversion", SpecVersion.V1.toString())
+                    .add("ce-id", ID)
+                    .add("ce-type", TYPE)
+                    .add("ce-source", SOURCE.toString())
+                    .add("ce-dataschema", DATASCHEMA.toString())
+                    .add("content-type", DATACONTENTTYPE_JSON)
+                    .add("ce-subject", SUBJECT)
+                    .add("ce-time", Time.writeTime(TIME))
+                    .add("ce-astring", "aaa")
+                    .add("ce-aboolean", "true")
+                    .add("ce-anumber", "10"),
+                DATA_JSON_SERIALIZED
+            ),
+            Arguments.of(
+                V1_WITH_XML_DATA,
+                createHeaders()
+                    .add("ce-specversion", SpecVersion.V1.toString())
+                    .add("ce-id", ID)
+                    .add("ce-type", TYPE)
+                    .add("ce-source", SOURCE.toString())
+                    .add("content-type", DATACONTENTTYPE_XML)
+                    .add("ce-subject", SUBJECT)
+                    .add("ce-time", Time.writeTime(TIME)),
+                DATA_XML_SERIALIZED
+            ),
+            Arguments.of(
+                V1_WITH_TEXT_DATA,
+                createHeaders()
+                    .add("ce-specversion", SpecVersion.V1.toString())
+                    .add("ce-id", ID)
+                    .add("ce-type", TYPE)
+                    .add("ce-source", SOURCE.toString())
+                    .add("content-type", DATACONTENTTYPE_TEXT)
+                    .add("ce-subject", SUBJECT)
+                    .add("ce-time", Time.writeTime(TIME)),
+                DATA_TEXT_SERIALIZED
+            )
+        );
+    }
+
+    private static Headers createHeaders() {
+        return new Headers();
+    }
+
+    private static class Headers extends HashMap<String, String> {
+        public Headers add(String k, String v) {
+            this.put(k, v);
+            return this;
+        }
+    }
+}
--- a/examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/BasicHttpServer.java
+++ b/examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/BasicHttpServer.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2018-Present The CloudEvents Authors
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cloudevents.examples.http.basic;
+
+import com.sun.net.httpserver.Headers;
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpServer;
+import io.cloudevents.CloudEvent;
+import io.cloudevents.core.message.MessageReader;
+import io.cloudevents.core.message.MessageWriter;
+import io.cloudevents.http.HttpMessageFactory;
+
+import java.io.*;
+import java.net.InetSocketAddress;
+
+public class BasicHttpServer {
+
+    public static void main(String[] args) throws IOException {
+        HttpServer httpServer = HttpServer.create(new InetSocketAddress("localhost", 8080), 0);
+        httpServer.createContext("/echo", BasicHttpServer::echoHandler);
+        httpServer.start();
+    }
+
+    private static void echoHandler(HttpExchange exchange) throws IOException {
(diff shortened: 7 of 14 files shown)


http/restful-ws/src/main/java/io/cloudevents/http/restful/ws/impl/RestfulWSMessageWriter.java
```diff
@@ -20,67 +20,36 @@
...
-        return null;
+        super((name,value) -> httpHeaders.add(name,value),
+            body -> {
+            try {
+
+            } finally {
+                try {
+                    if (body != null) {
+                        entityStream.write(body);
+                    }
```
matejvasek, line 47: Obviously this should be in the outer `try`.

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,118 @@
...
+    public static Wire writeCloudEventStructured(CloudEvent cloudEvent, String eventFormat) {
+        Wire result = new Wire();
+        Map<String,String> headers = new HashMap<>(4);
+        result.setHeaders(headers);
+        MessageWriter writer = createWriter(headers::put, bytes -> result.setBody(bytes));
+        writer.writeStructured(cloudEvent, eventFormat);
+        return result;
+    }
+
+    public static final class Wire {
```
matejvasek, line 98: Do we want this `Wire` stuff?

matejvasek:
relates to #212

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,118 @@
...
+            sv -> new HttpMessageReader(sv, headers.entrySet(), body),
+            UnknownEncodingMessageReader::new
+        );
+    }
+
+    public static MessageWriter createWriter(BiConsumer<String, String> putHeader, Consumer<byte[]> putBody) {
+        return new HttpMessageWriter(putHeader, putBody);
+    }
+
+    public static CloudEvent readCloudEvent(Map<String,String> headers, byte[] body) {
```
matejvasek, line 48: @eamonnmcmanus does this API makes sense to you?

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,118 @@
...
+        return createReader(headers, body).toEvent();
+    }
+
+    public static CloudEvent readCloudEvent(Wire wire) {
+        return createReader(wire.getHeaders(), wire.getBody()).toEvent();
+    }
+
+    public static void writeCloudEventBinary(CloudEvent cloudEvent, BiConsumer<String, String> putHeader, Consumer<byte[]> putBody) {
```
matejvasek, line 56: @eamonnmcmanus does this API makes sense to you?

http/restful-ws/src/main/java/io/cloudevents/http/restful/ws/impl/RestfulWSMessageWriter.java
```diff
@@ -20,67 +20,36 @@
 import io.cloudevents.SpecVersion;
 import io.cloudevents.core.format.EventFormat;
 import io.cloudevents.core.message.MessageWriter;
+import io.cloudevents.http.HttpMessageWriter;
 import io.cloudevents.rw.CloudEventRWException;
 import io.cloudevents.rw.CloudEventWriter;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
```
slinkydeveloper, line 26: What?!

slinkydeveloper (commented):
I saw you did some refactoring of restful-ws impl inside this PR, can you push it in a separate PR?

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
slinkydeveloper, line 48: Do we really need `readCloudEvent`/`writeCloudEvent`? The user goes from/to event with another single method invocation...

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,118 @@
...
+        return MessageUtils.parseStructuredOrBinaryMessage(
+            () -> headers.get("Content-Type"),
+            format -> new GenericStructuredMessageReader(format, body),
+            () -> headers.get(CloudEventsHeaders.SPEC_VERSION),
+            sv -> new HttpMessageReader(sv, headers.entrySet(), body),
+            UnknownEncodingMessageReader::new
+        );
+    }
+
+    public static MessageWriter createWriter(BiConsumer<String, String> putHeader, Consumer<byte[]> putBody) {
```
slinkydeveloper, line 44: That's a cool idea!

matejvasek:
Sorry for mixing up things:
Re:  restful-ws changes -- commit `just an example of usage` it there only to demonstrate usage, it is not supposed to be merged.

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
slinkydeveloper: Solve the previous comment about read/write CloudEvent, then we'll find out if we still need this

http/restful-ws/src/main/java/io/cloudevents/http/restful/ws/impl/RestfulWSMessageWriter.java
matejvasek: Should have been UnsupportedOperationException, I wrote NotImplementedException (because I used that in c# :smile:), then IDE did rest.

matejvasek: Anyway this is only for demonstration purposes only in while this is draft.

slinkydeveloper:
> @slinkydeveloper Sorry for mixing up things:
> Re: restful-ws changes -- commit `just an example of usage` it there only to demonstrate usage, it is not supposed to be merged.
> BTW it also shows that generic interface is not most efficient since I had to convert Multi Map to ordinary map iterator.

For demonstration purpose, can you add a snippet to the module readme? (also with the tests, the usage should be pretty clear)

http/basic/src/main/java/io/cloudevents/http/HttpMessageReader.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Map;
+import java.util.function.BiConsumer;
+
+import static io.cloudevents.http.CloudEventsHeaders.CE_PREFIX;
+
+public class HttpMessageReader extends BaseGenericBinaryMessageReaderImpl<String, String> {
+
+    private final Iterable<Map.Entry<String,String>> headers;
+
+    protected HttpMessageReader(SpecVersion version, Iterable<Map.Entry<String,String>> headers, byte[] body) {
```
eamonnmcmanus, line 32: What about just `Map<String, String>` for the `headers` parameter? I think that would be friendlier for users. Obviously you can make the type here out of a `Map<String, String>` by calling `map.entrySet()` but that feels a bit more complicated than it needs to be.

http/basic/src/main/java/io/cloudevents/http/HttpMessageWriter.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public Void end() {
+        putBody.accept(null);
+        return null;
+    }
+
+    @Override
+    public void setAttribute(String name, String value) throws CloudEventRWException {
+        System.err.println("name: " + name + ", value: " + value);
```
eamonnmcmanus, line 59: I know this is a draft but you could probably delete this. :-)

http/restful-ws/src/main/java/io/cloudevents/http/restful/ws/CloudEventsProvider.java
```diff
@@ -99,6 +99,7 @@ public void writeTo(CloudEvent event, Class<?> type, Type genericType, Annotatio
             .map(a -> ((StructuredEncoding) a).value())
             .findFirst();
 
+        httpHeaders.remove("Content-Type");
```
eamonnmcmanus, line 102: I'm not sure how this relates?

http/restful-ws/src/main/java/io/cloudevents/http/restful/ws/impl/BinaryRestfulWSMessageReaderImpl.java
```diff
@@ -19,47 +19,53 @@
...
+import java.util.*;
 import java.util.function.BiConsumer;
 
 import static io.cloudevents.http.restful.ws.impl.CloudEventsHeaders.CE_PREFIX;
 
-public final class BinaryRestfulWSMessageReaderImpl extends BaseGenericBinaryMessageReaderImpl<String, String> {
+public final class BinaryRestfulWSMessageReaderImpl extends HttpMessageReader {
 
-    private final MultivaluedMap<String, String> headers;
+    static private Iterable<Map.Entry<String, String>> multivaluedMapToIterableHeaders(MultivaluedMap<String, String> multivaluedMap) {
```
eamonnmcmanus, line 34: If the type of the `HttpMessageReader` constructor parameter were `Map<String, String>` then you could use something like this here:

```
multivaluedMap.entrySet().stream().collect(toMap(k -> k, v -> v.getFirst()));
```

(You could also use that with the existing type and then call `entrySet()` on it.)

http/restful-ws/src/main/java/io/cloudevents/http/restful/ws/impl/BinaryRestfulWSMessageReaderImpl.java
```diff
@@ -19,47 +19,53 @@
...
-    }
-
-    @Override
-    protected boolean isContentTypeHeader(String key) {
-        return key.equalsIgnoreCase(HttpHeaders.CONTENT_TYPE);
-    }
+                @Override
+                public Map.Entry<String, String> next() {
+                    Map.Entry<String, List<String>> n = it.next();
+                    return new Map.Entry<String, String>() {
```
eamonnmcmanus, line 46: FWIW `AbstractMap.SimpleImmutableEntry` would be simpler.

http/basic/src/main/java/io/cloudevents/http/HttpMessageReader.java
```diff
@@ -0,0 +1,61 @@
...
+        return key.length() > 3 && key.substring(0, CE_PREFIX.length()).toLowerCase().startsWith(CE_PREFIX);
+    }
+
+    @Override
+    protected String toCloudEventsKey(String key) {
+        return key.substring(CE_PREFIX.length()).toLowerCase();
+    }
+
+    @Override
+    protected void forEachHeader(BiConsumer<String, String> fn) {
```
eamonnmcmanus, line 53: If the type of `headers` is `Map<String, String>` then this method is just `headers.forEach(fn)`.

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,118 @@
...
+import io.cloudevents.core.message.impl.UnknownEncodingMessageReader;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+
+public class HttpMessageFactory {
+
+    public static MessageReader createReader(Map<String,String> headers, byte[] body) throws IllegalArgumentException {
```
matejvasek, line 34: @slinkydeveloper @eamonnmcmanus what about changing this to`Map<String, List<String>>`?

http/restful-ws/src/main/java/io/cloudevents/http/restful/ws/CloudEventsProvider.java
matejvasek: I am going to completely remove those changes, to explain: the header is set to text/html by default, but we don't won't that there.

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
eamonnmcmanus: That's technically more correct, since HTTP headers can be repeated. It would also allow you to pass in JAX-RS's `MultivaluedMap` directly. On the other hand, the CloudEvents spec doesn't use any repeated headers, and it's a bit less convenient to synthesize a `Map<String, List<String>>` in tests. But I think either type is defensible.

matejvasek: Well `java.net.HttpURLConnection` and `com.sun.net.httpserver.HttpServer` also use `Map<String, List<String>>`.

slinkydeveloper: I'm against this, because in the cloudevents spec the assumption is that there is one value for each key, so that doesn't really map to the cloudevents spec itself (although, for http, it makes sense)

slinkydeveloper: Maybe we can have an override that takes `Map<String, List<String>>` and translates it to `Map<String, String>`

eamonnmcmanus: That's a fair point. Tests can always use a trivial helper method that makes a `Map<String, List<String>>` out of a `Map<String, String>`:

```
private static Map<String, List<String>> toMultimap(Map<String, String> map) {
  return map.entrySet().stream().collect(toMap(e -> e.getKey(), e -> List.of(e.getValue())));
}
```

I'd suggest having both via overloads, but erasure would get in the way. So `Map<String, List<String>>` is probably better if you have to choose.

matejvasek: Would method overload via generics even work in Java? Or do you mean it would have slightly different name?

matejvasek:
I added examples using `java.net.HttpURLConnection` and `com.sun.net.httpserver.HttpServer`.

matejvasek:
Speaking of header: I must not forget about their case-insensitivity.

slinkydeveloper:
> Speaking of header: I must not forget about their case-insensitivity.

On the write side, i guess you should just take a convention and use it... Then the internal representation is up to the single http libraries

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,48 @@
...
+        return MessageUtils.parseStructuredOrBinaryMessage(
+            () -> headers.get("Content-Type"),
```
eamonnmcmanus, line 36: I guess this line and the one two below are where case-sensitivity comes into play? You could write a small helper method:

```
  private static <V> V getIgnoringCase(Map<String, V> map, String key) {
    for (Map.Entry<String, V> entry : map.entrySet()) {
      if (key.equalsIgnoreCase(entry.getKey())) {
        return entry.getValue();
      }
    }
  }
```

Or, you could copy the input map into a `new TreeMap<>(String.CASE_INSENSITIVE_ORDER)`.

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
slinkydeveloper: You can encapsulate in a `Wire` instance (name tbd) the actual headers and payload, then to create the wire you have methods like:

```
Wire of(Map<String, String>, byte[])
Wire ofMultimap(Map<String, String>, byte[])
```

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
matejvasek: Indeed it looks like this is the only place where capitalization matters. I don't like idea of creating `new` cases insensitive map just to do two `gets()`. That linear search might be better. I could do following:
```java
    public static MessageReader createReader(Map<String,String> headers, byte[] body) throws IllegalArgumentException {
        String _ct = null;
        String _sv = null;

        for (Map.Entry<String,String> kvp: headers.entrySet()) {
            if (kvp.getKey().equalsIgnoreCase(CONTENT_TYPE)) {
                _ct = kvp.getValue();
            } else if (kvp.getKey().equalsIgnoreCase(CloudEventsHeaders.SPEC_VERSION)) {
                _sv = kvp.getValue();
            }
        }

        final String specversion = _sv;
        final String contentType = _ct;

        return MessageUtils.parseStructuredOrBinaryMessage(
            () -> contentType,
            format -> new GenericStructuredMessageReader(format, body),
            () -> specversion,
            sv -> new HttpMessageReader(sv, headers, body),
            UnknownEncodingMessageReader::new
        );
    }
```

matejvasek: Other option would be delegate that to caller. Javadoc could state that the map has to be case insensitive, but I don't think that is good idea as users might miss that. wdyt @slinkydeveloper

slinkydeveloper: > Other option would be delegate that to caller. Javadoc could state that the map has to be case insensitive, but I don't think that is good idea as users might miss that. wdyt @slinkydeveloper

I think we should provide a case insensitive solution by ourselves, at least at the beginning. This module should be a best-effort but correct solution, not an optimized one: if you want the optimized solution, implement the reader/writer for your specific data structure, that already solves these problems by itself.

matejvasek: If I am going to use the linear search I could as well not use `java.util.Map`, but `Consumer<BiConsumer<String,String>>`
```java
public static MessageReader createReader(Consumer<BiConsumer<String,String>> forEachHeader, byte[] body) throws IllegalArgumentException {
        final String[] _ct = {null};
        final String[] _sv = {null};

        forEachHeader.accept((k, v) -> {
            if (k.equalsIgnoreCase(CONTENT_TYPE)) {
                _ct[0] = v;
            } else if (k.equalsIgnoreCase(CloudEventsHeaders.SPEC_VERSION)) {
                _sv[0] = v;
            }
        });

        final String specversion = _sv[0];
        final String contentType = _ct[0];

        return MessageUtils.parseStructuredOrBinaryMessage(
            () -> contentType,
            format -> new GenericStructuredMessageReader(format, body),
            () -> specversion,
            sv -> new HttpMessageReader(sv, forEachHeader, body),
            UnknownEncodingMessageReader::new
        );
    }
```

matejvasek: What's more desirable?
```java
public static MessageReader createReader(Map<String,String> headers, byte[] body) throws IllegalArgumentException;
// VS
public static MessageReader createReader(Consumer<BiConsumer<String,String>> forEachHeader, byte[] body) throws IllegalArgumentException;

Comment (slinkydeveloper) [reply http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java]: I think that biconsumer might suit well for multimaps too

Comment (slinkydeveloper) [reply http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java]: And we can have the `Map` signature as well

File: http/basic/src/test/java/io/cloudevents/http/impl/HttpMessageWriterTest.java [hunk 15]
```diff
@@ -0,0 +1,225 @@
...
+
+        expectedHeaders.forEach((k, v) -> {
+            assertThat(headers.get(k)).isEqualTo(v);
+        });
+        if (expectedBody != null) {
+            assertThat(body[0]).isEqualTo(expectedBody);
+        }
+    }
+
+    public static Stream<Arguments> binaryTestArguments() {
```
matejvasek, line 80: This code is already copypasted with little changes at multiple places. We should refactor that latter (not in this PR) if possible.

slinkydeveloper: Maybe add a todo to the code itself?

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
eamonnmcmanus: The advantage of the `Consumer<BiConsumer<String, String>>` solution is that it sidesteps the question of whether to use `Map<String, String>` or `Map<String, List<String>>`, but the disadvantage is that it's kind of daunting for users. If I see a parameter `Consumer<BiConsumer<String, String>>` my first reaction is to panic. :-)
I think it would be OK, if the javadoc showed how to use it with the two `Map` cases, something like this:

```
Map<String, String> headers = ...;
MessageReader reader = HttpMessageFactory.createReader(
    forEach -> headers.forEach(forEach), body);
```
```
Map<String, List<String>> headers = ...;
Consumer<BiConsumer<String, String>> consumer = forEach ->
    headers.forEach((key, values) -> values.forEach(value -> forEach.accept(key, value)));
MessageReader reader = HttpMessageFactory.createReader(forEach, body);
```

As @slinkydeveloper suggests, we could also have both `Consumer` and `Map` overloads. The `Map` overload could have javadoc that says it's equivalent to one of the code extracts above (depending on what flavour of `Map` it is). Then the `Consumer` overload javadoc could show the other code extract.

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,82 @@
...
+import java.net.HttpURLConnection;
+import java.net.URLConnection;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+
+import static io.cloudevents.http.impl.CloudEventsHeaders.CONTENT_TYPE;
+
+public class HttpMessageFactory {
```
matejvasek, line 40: @slinkydeveloper @eamonnmcmanus I added requested methods, does it look good now?

matejvasek:
I added README and some examples `HttpURLConnection`, `Servlet API`, `com.sun.net.httpserver.HttpServer`.

matejvasek:
TODO: add Javadocs

matejvasek:
Should I call it `basic http` or `generic http`?

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public static MessageReader createReader(Consumer<BiConsumer<String,String>> forEachHeader, byte[] body) {
```
matejvasek, line 42: I am thinking of adding overload for `InputStream` body since many API use that.

matejvasek:
```java
public static MessageReader createReader(Consumer<BiConsumer<String,String>> forEachHeader, InputStream body);
```

eamonnmcmanus: I don't think that's worthwhile. Users can trivially convert an `InputStream` into a `byte[]` with [this method](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/InputStream.html#readAllBytes()).

matejvasek: The method is not in java8, is it? Another option would be call IOUtils from apache commons, but that is yet another dependency.

matejvasek: On the other hand `byte[]` is trivially convertible to `InputStream` by `ByteArrayInputStream` which is in java since `1.0`.

eamonnmcmanus: Java 8 is End Of Life. I don't think new APIs should be making design decisions based on what it does or does not have. Anyway, even users who need to stay on Java 8 will not find it hard to make a `byte[]` if they have an `InputStream`.

On the other hand, if the API had _only_ `InputStream` rather than `byte[]`, I think that would be fine. As you point out, users can convert from the latter to the former trivially. And in the typical context of an HTTP server (servlet, or whatever) you typically do get the HTTP request body as an `InputStream`.

matejvasek: > And in the typical context of an HTTP server (servlet, or whatever) you typically do get the HTTP request body as an InputStream.

That's why I am thinking of it. I tried to create 3 examples, all use InputStream.

slinkydeveloper:
For me the name of the module should be `cloudevents-http-basic`

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
slinkydeveloper: I want to keep everything consistent (as much as we can) and use a single type for the bytes. For now we use `byte[]` and i think here you should have only the `byte[]` signature, if users need it, we add the overload later for `InputStream`. But please keep everything `byte[]` based atm.
I've opened another issue for considering switchin byte[] to ByteBuffer all over the sdk https://github.com/cloudevents/sdk-java/issues/207

examples/basic-http/pom.xml
```diff
@@ -44,6 +44,11 @@
             <version>${project.version}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
slinkydeveloper, line 47: Nope :smile:

matejvasek:
@eamonnmcmanus pleas give it another look.

http/basic/README.md
```diff
@@ -0,0 +1,189 @@
...
+</dependency>
+```
+
+## Sending and Receiving CloudEvents
+To send and receive CloudEvents we use `MessageWriter` and `MessageReader`, respectively.
+This module offers generic factory methods for creation of those in `HttpMessageFactory`.
+
+```java
+public class HttpMessageFactory {
+    public static MessageReader createReader(Consumer<BiConsumer<String,String>> forEachHeader, byte[] body);
```
matejvasek, line 21: Out of date with code.

http/basic/src/test/java/io/cloudevents/http/HttpMessageReaderWriterTest.java
```diff
@@ -0,0 +1,255 @@
...
+                    .add("ce-source", SOURCE.toString())
+                    .add("content-type", DATACONTENTTYPE_TEXT)
+                    .add("ce-subject", SUBJECT)
+                    .add("ce-time", Time.RFC3339_DATE_FORMAT.format(TIME)),
+                DATA_TEXT_SERIALIZED
+            )
+        );
+    }
+
+    private static Headers createHeaders() {
```
matejvasek, line 241: This could just return `HashMap`.

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/HttpURLConnectionClient.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private static MessageWriter getWriteMessage(HttpURLConnection httpUrlConnection) {
+        return HttpMessageFactory.createWriter(
+            httpUrlConnection::setRequestProperty,
+            body -> {
+                try {
+                    if (body != null) {
+                        httpUrlConnection.setRequestProperty("content-length", String.valueOf(body.length));
+                        OutputStream outputStream = httpUrlConnection.getOutputStream();
```
eamonnmcmanus, line 87:
```
try (OutputStream outputStream = httpUrlConnection.getOutputStream()) {
    outputStream.write(body);
}
```

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/IOUtils.java
```diff
@@ -0,0 +1,56 @@
...
+// based on apache commons io
+public class IOUtils {
+
+    private static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
+
+    public static byte[] toByteArray(InputStream body, int size) throws IOException {
+        if (size < 0) throw new IllegalArgumentException("size cannot be negative.");
+        if (size > 0) {
+            byte[] data = new byte[size];
+            int offset = 0;
```
eamonnmcmanus, line 32:
```
new DataInputStream(body).readFully(data);
```

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/JettyServer.java
```diff
@@ -0,0 +1,111 @@
...
+            Enumeration<String> headerNames = httpServletRequest.getHeaderNames();
+            while (headerNames.hasMoreElements()) {
+                String name = headerNames.nextElement();
+                processHeader.accept(name, httpServletRequest.getHeader(name));
+
+            }
+        };
+
+        byte[] body;
+        String contentLength = httpServletRequest.getHeader("Content-Length");
```
eamonnmcmanus, line 75:
```
int contentLength = httpServletRequest.getContentLength();
if (contentLength >= 0) {
```

http/basic/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+    public static MessageWriter messageWriterFromResponse(HttpServletResponse httpServletResponse) throws IOException {
+        return HttpMessageFactory.createWriter(
+            httpServletResponse::addHeader,
+            body -> {
+                try {
+                    if (body != null) {
+                        httpServletResponse.setContentLength(body.length);
+                        httpServletResponse.setStatus(HttpStatus.OK_200);
+                        ServletOutputStream outputStream = httpServletResponse.getOutputStream();
```
eamonnmcmanus, line 54: try-with-resources

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,166 @@
...
+    private HttpMessageFactory() {}
+
+    /**
+     * Creates a new {@link MessageReader} that can read both structured and binary messages from a HTTP response (client) or request (server).
+     *
+     * <pre>
+     * Example of usage with <a href="https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletRequest.html">HttpServletRequest</a>:
+     * {@code
+     * Consumer<BiConsumer<String,String>> forEachHeader = processHeader -> {
+     *     Enumeration<String> headerNames = httpServletRequest.getHeaderNames();
```
eamonnmcmanus, line 55:
```
for (String name : Collections.list(httpServletRequest.getHeaderNames())) {
```

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,166 @@
...
+     *     while (headerNames.hasMoreElements()) {
+     *         String name = headerNames.nextElement();
+     *         processHeader.accept(name, httpServletRequest.getHeader(name));
+     *
+     *     }
+     * };
+     * byte[] body;
```
eamonnmcmanus, line 62: I think you could use `InputStream.readAllBytes()` in this example. People who are stuck on Java 8 can figure out how to achieve the same thing. I also don't think there's any harm in ignoring the `Content-Length`.

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,166 @@
...
+     * HttpMessageFactory.createReader(forEachHeader, body);
+     * }
+     * </pre>
+     * @param forEachHeader http headers visitor function
+     * @param body          nullable buffer of the body
+     * @return a message reader implementation with potentially an unknown encoding
+     * @throws IllegalArgumentException If, in case of binary mode, the spec version is invalid
+     */
+    public static MessageReader createReader(Consumer<BiConsumer<String, String>> forEachHeader, byte[] body) {
+        final String[] hdrs = {null, null};
```
eamonnmcmanus, line 78: There's no really good solution here, but I think it would be better to have
```
AtomicReference<String> contentType = new AtomicReference<>();
AtomicReference<String> specVersion = new AtomicReference<>();
```
and then use `contentType.get()` and `specVersion.get()` directly in the call to `parseStructuredOrBinaryMessage`.

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Creates a new {@link MessageReader} that can read both structured and binary messages from a HTTP response (client) or request (server).
+     *
+     * @param headers http headers as multimap
+     * @param body    nullable buffer of the body
+     * @return a message reader implementation with potentially an unknown encoding
+     * @throws IllegalArgumentException If, in case of binary mode, the spec version is invalid
+     */
+    public static MessageReader createReaderFromMultiMap(Map<String, List<String>> headers, byte[] body) {
```
eamonnmcmanus, line 124: I think the spelling Multimap rather than MultiMap is more usual.

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,166 @@
...
+     * Example of usage with <a href="https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletResponse.html">HttpServletResponse</a>:
+     * {@code
+     * HttpMessageFactory.createWriter(
+     *     httpServletResponse::addHeader,
+     *     body -> {
+     *         try {
+     *             if (body != null) {
+     *                 httpServletResponse.setContentLength(body.length);
+     *                 httpServletResponse.setStatus(HttpStatus.OK_200);
+     *                 ServletOutputStream outputStream = httpServletResponse.getOutputStream();
```
eamonnmcmanus, line 146: try-with-resources

http/basic/src/main/java/io/cloudevents/http/impl/CloudEventsHeaders.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package io.cloudevents.http.impl;
+
+import io.cloudevents.core.message.impl.MessageUtils;
+
+import java.util.Collections;
+import java.util.Map;
+
+public class CloudEventsHeaders {
```
eamonnmcmanus, line 24: I'd add a private constructor.

eamonnmcmanus (commented):
This is looking good. Thanks for putting in all that work!

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
matejvasek: Definitely would work, but why allocate new List just to iterate over it?

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
matejvasek: I was thinking of of create private static class with the two properties.

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/HttpServer.java
```diff
@@ -0,0 +1,81 @@
...
+            System.out.println("Handling event: " + cloudEvent);
+
+            MessageWriter messageWriter = getWriteMessage(exchange);
+            messageWriter.writeBinary(cloudEvent);
+        } catch (Throwable t) {
+            exchange.sendResponseHeaders(500, 0);
+        }
+    }
+
+    private static MessageReader getReadMessage(HttpExchange httpExchange) throws IOException {
```
slinkydeveloper, line 56: s/getReadMessage/createMessageReader

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/HttpServer.java
```diff
@@ -0,0 +1,81 @@
...
+        String contentLength = headers.getFirst("Content-Length");
+        if (contentLength != null) {
+            body = IOUtils.toByteArray(httpExchange.getRequestBody(), Integer.parseInt(contentLength));
+        } else {
+            body = IOUtils.toByteArray(httpExchange.getRequestBody());
+        }
+        return HttpMessageFactory.createReaderFromMultiMap(headers, body);
+    }
+
+    private static MessageWriter getWriteMessage(HttpExchange httpExchange) {
```
slinkydeveloper, line 68: s/getWriteMessage/createMessageWriter

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/HttpURLConnectionClient.java
```diff
@@ -0,0 +1,98 @@
...
+        messageWriter.writeBinary(ceToSend);
+
+        MessageReader messageReader = getReadMessage(httpUrlConnection);
+        CloudEvent receivedCE = messageReader.toEvent();
+
+        System.out.println("CloudEvent: " + receivedCE);
+        System.out.println("Data: " + new String(receivedCE.getData(), StandardCharsets.UTF_8));
+    }
+
+    private static MessageReader getReadMessage(HttpURLConnection httpUrlConnection) throws IOException {
```
slinkydeveloper, line 63: same as above

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/HttpURLConnectionClient.java
```diff
@@ -0,0 +1,98 @@
...
+        if (contentLength != null) {
+            body = IOUtils.toByteArray(httpUrlConnection.getInputStream(), Integer.parseInt(contentLength));
+        } else {
+            body = IOUtils.toByteArray(httpUrlConnection.getInputStream());
+        }
+
+        return HttpMessageFactory.createReaderFromMultiMap(headers, body);
+    }
+
+    private static MessageWriter getWriteMessage(HttpURLConnection httpUrlConnection) {
```
slinkydeveloper, line 80: same as above

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/IOUtils.java
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+package io.cloudeventrs.examples.http;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+// based on apache commons io
```
slinkydeveloper, line 23: add a link to the source code

slinkydeveloper (commented):
I did a first pass, on monday i'll do another one :smile:

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
slinkydeveloper: I prefer MultiMap because to me it's like Multimap, but it's a matter of taste

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
matejvasek: I think that with known size the byte array can be allocated straight away. With unknown size there could be reallocation. But who knows what optimizations are under the hood.

matejvasek: And yes, it's just an example, but then again some people are going to just copypaste this.

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/IOUtils.java
matejvasek: better now?

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/HttpServer.java
```diff
@@ -0,0 +1,81 @@
...
+            System.out.println("Handling event: " + cloudEvent);
+
+            MessageWriter messageWriter = createWriteMessage(exchange);
+            messageWriter.writeBinary(cloudEvent);
+        } catch (Throwable t) {
+            exchange.sendResponseHeaders(500, 0);
+        }
+    }
+
+    private static MessageReader createReadMessage(HttpExchange httpExchange) throws IOException {
```
slinkydeveloper, line 56: The type is called `MessageReader`, so i think this should be named `createMessageReader` (same below with writers too)

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/HttpServer.java
```diff
@@ -0,0 +1,81 @@
...
+
+public class HttpServer {
+
+    public static void main(String[] args) throws IOException {
+        com.sun.net.httpserver.HttpServer httpServer = com.sun.net.httpserver.HttpServer.create(new InetSocketAddress("localhost", 8080), 0);
+        httpServer.createContext("/echo", HttpServer::handle);
+        httpServer.start();
+    }
+
+    public static void handle(HttpExchange exchange) throws IOException {
```
slinkydeveloper, line 38: maybe `echoHandler`?

slinkydeveloper: Can you add a bunch of comments (nothing particularly verbose), in particular when you use the sdk methods?

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/HttpServer.java
```diff
@@ -0,0 +1,81 @@
...
+        try {
+            MessageReader messageReader = createReadMessage(exchange);
+            CloudEvent cloudEvent = messageReader.toEvent();
+
+            System.out.println("Handling event: " + cloudEvent);
+
+            MessageWriter messageWriter = createWriteMessage(exchange);
+            messageWriter.writeBinary(cloudEvent);
+        } catch (Throwable t) {
+            exchange.sendResponseHeaders(500, 0);
```
slinkydeveloper, line 52: log the error here and/or return the error to the response as text?

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/HttpServer.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    private static MessageReader createReadMessage(HttpExchange httpExchange) throws IOException {
+        Headers headers = httpExchange.getRequestHeaders();
+        byte[] body;
+        String contentLength = headers.getFirst("Content-Length");
```
slinkydeveloper, line 59: That sounds like an optimization, can we remove it? We don't need to show it for the sdk usage

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/HttpServer.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private static MessageWriter createWriteMessage(HttpExchange httpExchange) {
+        return HttpMessageFactory.createWriter(
+            httpExchange.getResponseHeaders()::add,
+            body -> {
+                try {
+                    httpExchange.sendResponseHeaders(200, body.length);
+                    httpExchange.getResponseBody().write(body);
+                } catch (IOException t) {
+                    throw new UncheckedIOException(t);
```
slinkydeveloper, line 76: waaaaat, i didn't knew it :heart_eyes:

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/HttpURLConnectionClient.java
```diff
@@ -0,0 +1,98 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cloudeventrs.examples.http;
```
slinkydeveloper, line 17: typo in package name

slinkydeveloper: btw rename the package in `io.cloudevents.examples.http.basic`

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/HttpURLConnectionClient.java
```diff
@@ -0,0 +1,98 @@
...
+        System.out.println("CloudEvent: " + receivedCE);
+        System.out.println("Data: " + new String(receivedCE.getData(), StandardCharsets.UTF_8));
+    }
+
+    private static MessageReader createReadMessage(HttpURLConnection httpUrlConnection) throws IOException {
+
+        Map<String, List<String>> headers = httpUrlConnection.getHeaderFields();
+        byte[] body;
+        String contentLength = null;
+        if (headers.containsKey("Content-Length")) {
```
slinkydeveloper, line 68: same optimization as before, please remove it

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/IOUtils.java
```diff
@@ -0,0 +1,60 @@
...
+// based on apache's commons-io
+// which is licensed under the Apache License, Version 2.0
+// https://commons.apache.org/proper/commons-io/
+// https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/IOUtils.java
+
+public class IOUtils {
+
+    private static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
+
+    public static byte[] toByteArray(InputStream body, int size) throws IOException {
```
slinkydeveloper, line 32: You don't need this method if you remove the optimizations above

examples/basic-http/src/main/java/io/cloudeventrs/examples/http/JettyServer.java
```diff
@@ -0,0 +1,111 @@
...
+            Enumeration<String> headerNames = httpServletRequest.getHeaderNames();
+            while (headerNames.hasMoreElements()) {
+                String name = headerNames.nextElement();
+                processHeader.accept(name, httpServletRequest.getHeader(name));
+
+            }
+        };
+
+        byte[] body;
+        int contentLength = httpServletRequest.getContentLength();
```
slinkydeveloper, line 75: Same

http/basic/README.md
```diff
@@ -0,0 +1,209 @@
...
+    <artifactId>cloudevents-http-basic</artifactId>
+    <version>2.0.0-milestone1</version>
+</dependency>
+```
+
+## Sending and Receiving CloudEvents
+To send and receive CloudEvents we use `MessageWriter` and `MessageReader`, respectively.
+This module offers generic factory methods for creation of those in `HttpMessageFactory`.
+
+```java
```
slinkydeveloper, line 19: Remove this, link to the code

http/basic/README.md
```diff
@@ -0,0 +1,209 @@
...
+    public static MessageReader createReader(Consumer<BiConsumer<String,String>> forEachHeader, byte[] body);
+    public static MessageReader createReader(Map<String,String> headers, byte[] body);
+    public static MessageReader createReaderFromMultimap(Map<String,List<String>> headers, byte[] body);
+    public static MessageWriter createWriter(BiConsumer<String, String> putHeader, Consumer<byte[]> sendBody);
+}
+```
+
+Below you can see examples of usage.
+
+### HttpServlet's Request and Response
```
slinkydeveloper, line 30: Remove all this code below for this and other http impls and point to the sample

http/basic/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+```xml
+<dependency>
+    <groupId>io.cloudevents</groupId>
+    <artifactId>cloudevents-http-basic</artifactId>
+    <version>2.0.0-milestone1</version>
+</dependency>
+```
+
+## Sending and Receiving CloudEvents
```
slinkydeveloper, line 15: We need to add a paragraph explaining why and why not this module

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
```diff
@@ -0,0 +1,166 @@
...
+     * Consumer<BiConsumer<String,String>> forEachHeader = processHeader -> {
+     *     Enumeration<String> headerNames = httpServletRequest.getHeaderNames();
+     *     while (headerNames.hasMoreElements()) {
+     *         String name = headerNames.nextElement();
+     *         processHeader.accept(name, httpServletRequest.getHeader(name));
+     *
+     *     }
+     * };
+     * byte[] body;
+     * int contentLength = httpServletRequest.getContentLength();
```
slinkydeveloper, line 63: remove this

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
slinkydeveloper: Yep +1 for `AtomicReference`, the arrays are kinda weird. I think a class here is overkilling too

http/basic/src/test/java/io/cloudevents/http/HttpMessageReaderWriterTest.java
```diff
@@ -0,0 +1,255 @@
...
+import static io.cloudevents.core.test.Data.*;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class HttpMessageReaderWriterTest {
+
+    @ParameterizedTest
+    @MethodSource("io.cloudevents.core.test.Data#allEventsWithoutExtensions")
+    void testWriteStructured(CloudEvent event) {
+
+        final byte[][] body = {null};
```
slinkydeveloper, line 47: Can you use AtomicReference here too?

http/basic/src/test/java/io/cloudevents/http/HttpMessageReaderWriterTest.java
```diff
@@ -0,0 +1,255 @@
...
+                DATA_TEXT_SERIALIZED
+            )
+        );
+    }
+
+    private static Headers createHeaders() {
+        return new Headers();
+    }
+
+    private static class Headers extends HashMap<String, String> {
```
slinkydeveloper, line 245: Why that?

slinkydeveloper (commented):
Because the basic http example contains examples with sun http server, http url connection and jetty server, maybe you should add a basic readme that explains the three different samples here?

http/basic/README.md
matejvasek: How should I reference the code? Hyperlink to github?

http/basic/src/test/java/io/cloudevents/http/HttpMessageReaderWriterTest.java
matejvasek: Will do. BTW since java10 it could be
```java
var bodyWrapper = new Object(){ byte[] value; };
```

http/basic/README.md
slinkydeveloper: yep just relative link to code, check out how to do that in other readmes

http/basic/README.md
```diff
@@ -0,0 +1,38 @@
+# Generic HTTP Protocol Binding
+
+[![Javadocs](http://www.javadoc.io/badge/io.cloudevents/cloudevents-http-basic.svg?color=green)](http://www.javadoc.io/doc/io.cloudevents/cloudevents-http-basic)
+
+This module is designed to be usable with various HTTP APIs.
```
matejvasek, line 5: @slinkydeveloper How should I improve this?

matejvasek:
@eamonnmcmanus please give it another look.

http/basic/README.md
slinkydeveloper: sounds ok to me, I would prefer you add links to other http impls (not only vertx, but also restful-ws)

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
eamonnmcmanus: A `TreeMap` is just a kind of `Map`, but a `Multimap` is a different sort of thing. As words we would write "tree map" but "multimap". https://en.wikipedia.org/wiki/Multimap

examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/HttpServer.java
```diff
@@ -0,0 +1,89 @@
...
+import com.sun.net.httpserver.HttpExchange;
+import io.cloudevents.CloudEvent;
+import io.cloudevents.core.message.MessageReader;
+import io.cloudevents.core.message.MessageWriter;
+import io.cloudevents.http.HttpMessageFactory;
+
+import java.io.*;
+import java.net.InetSocketAddress;
+
+public class HttpServer {
```
eamonnmcmanus, line 29: Maybe call this `ExampleHttpServer` or `BasicHttpServer` to avoid the clash with `com.sun.net.httpserver.HttpServer`?

examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/IOUtils.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.InputStream;
+
+// based on apache's commons-io
+// which is licensed under the Apache License, Version 2.0
+// https://commons.apache.org/proper/commons-io/
+// https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/IOUtils.java
+
+public class IOUtils {
+
+    public static byte[] toByteArray(InputStream body) throws IOException {
```
eamonnmcmanus, line 30: Maybe add a comment here saying that in Java 9+ you can just use `InputStream.getAllBytes()`.

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
eamonnmcmanus: `Collections.list` was introduced specifically to get the legacy type `Enumeration` out of the way as quickly as possible. For me this is the idiomatic way to iterate over the `Enumeration<String>` type that the ancient Servlet API imposes. It's also a lot more readable for the example code.

http/basic/src/test/java/io/cloudevents/http/HttpMessageReaderWriterTest.java
```diff
@@ -0,0 +1,247 @@
...
+        assertThat(message.toEvent())
+            .isEqualTo(expectedEvent);
+    }
+
+    public static Stream<Arguments> binaryTestArguments() {
+        return Stream.of(
+            // V03
+            Arguments.of(
+                V03_MIN,
+                new HashMap<String,String>() {{
```
eamonnmcmanus, line 112: [Every time you use double-brace initialization, God kills a kitten](https://blog.jooq.org/2014/12/08/dont-be-clever-the-double-curly-braces-anti-pattern/)

Probably OK in a test, but you could also write a helper function that takes a `String[]` or a `String[][]` and makes a `Map` out of it. Better still would be to use `Map.of(...)`, but that requires Java 9+.

matejvasek: Originally I used something different: https://github.com/cloudevents/sdk-java/pull/225/commits/c09ef8c7f400a2ce97ef1de19f42f06ec4ab120c#diff-da38f4f0dad4b560cd509239f96a762eL241 was it better?

matejvasek: I didn't realize it uses anon classes. Gee couldn't they just already add proper literals to the language.

http/basic/src/main/java/io/cloudevents/http/HttpMessageFactory.java
matejvasek: Just found out that there is `asIterator()` then I could call `forEachRemaining` unfortunately the `asIterator()` method is available since Java 9+ :disappointed:

matejvasek: I agree that
```java
for (String name : Collections.list(httpServletRequest.getHeaderNames())) {}
```
is probably more readable.

But I find `ArrayList` allocation (and possible internal reallocation on its growth) unnecessary.
I specifically chosen that obsolete type to show that it works well with `Consumer<BiConsumer<String,String>> forEachHeader`. @slinkydeveloper wdyt?

http/basic/src/test/java/io/cloudevents/http/HttpMessageReaderWriterTest.java
eamonnmcmanus: Of course this is only a test, so the kittens God is killing are only test kittens. Still, I think I preferred the way you had it before, which was at least as readable without triggering the kitten-killing fear in seasoned Java developers.

For literals, I think `List.of(...)` and `Set.of(...)`, added in Java 9, are good enough for those types without needing a special syntax. There's also `Map.of(...)`, which you could have used here except for the desire to work on Java 8. But that one is a bit dicey anyway since there's no obvious syntactic difference between keys and values, and it arbitrarily stops at 10 key-value pairs.

http/basic/src/test/java/io/cloudevents/http/HttpMessageReaderWriterTest.java
matejvasek: I reverted that double brace init.

slinkydeveloper:
@matejvasek can you fix the dco?

slinkydeveloper (commented):
last change and we're ready to go

examples/basic-http/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-http-basic</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-json-jackson</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
matejvasek: also should be in `cloudevents-http-basic`

slinkydeveloper:
Thanks @matejvasek for the great work and thanks to @eamonnmcmanus for the fantastic job with the reviews!

PR #226, opened 2020-09-03
Add Quarkus example

ruromero:
Adding a Quarkus example. It exposes an endpoint to consume CloudEvents and periodically emits CloudEvents to this same endpoint using a REST client.

See the README for more details.

Signed-off-by: ruromero <rromerom@redhat.com>


matejvasek:
I am pleasantly surprised this work also in native mode. The reason I am surprised is because I would expected the `User` class cause troubles due to reflection.

Some time ago I tried something similar to this, to make it work I had to use `@RegisterForReflection`, otherwise I got errors.

Have Quarkus team done some magic to fix that in newer versions?

slinkydeveloper (commented):
Sounds good but I wonder if we can strip down a little bit more the sample:

* Can we remove the dockerfiles? Or do we really need these?
* Do we need all the stuff in the pom or can we simplify it?

Remember, the goal of this sample is to show how simple is to use sdk-java + restful-ws integration inside quarkus, not how to build a full app using quarkus. I think you could even link in the readme your example, showing how to build a full app, but i want to keep this one simple :smile: 

Also, because you're showing the restful-ws integration, i think you should rename the sample `restful-ws-quarkus`

ruromero:
> Has Quarkus team done some magic to fix that in newer versions?

This is what I was expecting and actually I initially added the annotation and later remove it just to try out. So maybe something has changed indeed. I will ask around and come back to you with the proper answer.

>     * Can we remove the dockerfiles? Or do we really need these?
Ok, I'll take out everything which is not directly related to the application. I thought leaving everything which is generated from the Quarkus Maven Plugin would be better for users getting started with Quarkus.
>     * Do we need all the stuff in the pom or can we simplify it?
If a user wants to try out that the native build works as well I would keep the pom as is. Native builds are very important in Quarkus as you saw in the previous comment.

> Also, because you're showing the restful-ws integration, i think you should rename the sample `restful-ws-quarkus`

Makes sense, I'll do it.

slinkydeveloper:
> I thought leaving everything which is generated from the Quarkus Maven Plugin would be better for users getting started with Quarkus.

That's my point, this is a getting started for sdk-java to use it in quarkus, not a getting started with quarkus, that's why I also suggest you to link to a more "proper example" of a full application, because you can then show all the features of using quarkus and integrating properly with cloudevents. This IMO must be a "start using cloudevents sdk in quarkus" and that's all.

ruromero:
Ready for another review

slinkydeveloper (approved):

slinkydeveloper:
Thanks @ruromero for the contribution!

slinkydeveloper:
Hey @ruromero this PR fails on master CI when tested with GraalVM, can you please check it out why? https://travis-ci.org/github/cloudevents/sdk-java/builds/724844933

matejvasek:
@slinkydeveloper I tried this on my machine with GraalVM:

openjdk version "11.0.7" 2020-04-14
OpenJDK Runtime Environment GraalVM CE 20.1.0 (build 11.0.7+10-jvmci-20.1-b02)
OpenJDK 64-Bit Server VM GraalVM CE 20.1.0 (build 11.0.7+10-jvmci-20.1-b02, mixed mode, sharing)

and test passes.

matejvasek:
https://github.com/quarkusio/quarkus/issues/11433 might be releated

ruromero:
I did all the builds and tests with GraalVM 2.1.0

matejvasek:
It might be something quarkus + custom maven setting.xml as in the aforementioned issue.

slinkydeveloper:
Maybe we need to update to latest graalvm release? Any ideas of how we could fix it?

ruromero:
I have created this PR https://github.com/cloudevents/sdk-java/pull/230

PR #232, opened 2020-09-11
Fix Graalvm travis build

ruromero:
Signed-off-by: ruromero <rromerom@redhat.com>

The Quarkus tests will require the other cloudevents libraries to be in the local maven repository, that's why the goal has been changed from `verify` to `install`


matejvasek:
Maybe it should be `mvn clean install -DskipTests=true -Dmaven.javadoc.skip=true && mvn verify` ?

PR #234, opened 2020-09-15
Add ARM64 support for Travis-CI

odidev:
Added ARM64 job in Travis-CI for openjdk8, openjdk11 and graalvm openjdk11.

Signed-off-by: odidev <odidev@puresoftware.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,12 @@ jobs:
     - name: "OpenJDK 11"
       jdk: openjdk11
       script: mvn clean verify -B
-    # GraalVM
+    - if: branches == master
+      name: "OpenJDK 11"
+      arch: arm64
+      jdk: openjdk11
+      script: ./mvnw clean verify -B
+   # GraalVM
     - name: "GraalVM 20.1.0 (8)"
       if: type != pull_request
       install: . ./install-jdk.sh --url "https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-20.1.0/graalvm-ce-java8-linux-amd64-20.1.0.tar.gz"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -22,10 +25,24 @@ jobs:
       name: "OpenJDK 8"
       jdk: openjdk8
       script: mvn clean verify -B
+    - stage: test
```
slinkydeveloper, line 28: Can we enable this only on master?

slinkydeveloper: Why stage:test here too?

.travis.yml
```diff
@@ -22,10 +25,24 @@ jobs:
...
+      addons:
+        apt:
+          - openjdk-8-jdk
+      jdk: openjdk8
+      script: mvn clean verify -B
     - name: "OpenJDK 11"
       jdk: openjdk11
       script: mvn clean verify -B
-    # GraalVM
+    - name: "OpenJDK 11"
```
slinkydeveloper, line 40: Can we enable this only on master?

.travis.yml
```diff
@@ -1,6 +1,9 @@
 language: java
 
 before_install:
+  - if [[ "${TRAVIS_CPU_ARCH}" == "arm64" ]]; then
```
slinkydeveloper, line 4: Why do we need to install maven from apt-get?

slinkydeveloper (commented):
May I ask why we need this? All the api/core modules doesn't depend on any transitive jni code

bsideup:
Just curious... What's the point of testing a pure Java library on ARM64 when JVM's ARM64 compatibility is already tested? Won't it add an unnecessary load to the CI for... nothing?

.travis.yml
bsideup: Especially given that there is a Maven Wrapper in the project that should be unused instead

odidev:
Although this is a pure java library but while i was testing it in travis i found out that travis is using graal vm for testing, these graalvm binaries mentioned can only be tested for amd64 platforms. I have updated the graalvm binaries link for the testing for arm64 platform.

.travis.yml
odidev: I have enable the openjdk tests only for master branch and removed any reductant tags (such as stage)

.travis.yml
odidev: I have enabled the openjdk arm64 tests for master only

.travis.yml
odidev: I have used the maven wrapper present in this.

slinkydeveloper:
@odidev the graalvm tests we do run in vm mode, they don't compile and test our application in native mode

slinkydeveloper:
This is not valid anymore because we migrated to gh actions

PR #249, opened 2020-09-29
Serialize time to RFC3339 compliant string

zhxnlai:
According to the [documentation](https://docs.oracle.com/javase/8/docs/api/java/time/OffsetDateTime.html#toString--), `OffsetDateTime`'s `toString` method uses this format `uuuu-MM-dd'T'HH:mmXXXXX` when seconds is zero. This is incompatible with RFC3339 and cannot be parsed by other cloud events SDKs. 

Here is an demo of the incompatibility:
https://play.golang.org/p/9db3zuJTM_k

### DIFF
--- a/api/src/main/java/io/cloudevents/types/Time.java
+++ b/api/src/main/java/io/cloudevents/types/Time.java
@@ -20,6 +20,8 @@ package io.cloudevents.types;
 import java.time.OffsetDateTime;
 import java.time.format.DateTimeParseException;
 
+import static java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME;
+
 /**
  * Utilities to handle the <a href="https://github.com/cloudevents/spec/blob/v1.0/spec.md#type-system">CloudEvent Attribute Timestamp type</a>
  */
@@ -36,9 +38,9 @@ public final class Time {
     }
 
     /**
-     * Convert a {@link OffsetDateTime} to {@link String}
+     * Convert a {@link OffsetDateTime} to a RFC3339 compliant {@link String}
      */
     public static String writeTime(OffsetDateTime time) throws DateTimeParseException {
-        return time.toString();
+        return ISO_OFFSET_DATE_TIME.format(time);
     }
 }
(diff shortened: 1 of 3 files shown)


matejvasek:
In general looks good, but it breaks some test since it adds unnecessary zeros to fraction part of date (`2018-04-26T14:48:09.123400Z` instead of `2018-04-26T14:48:09.1234Z`).
Maybe:
```java
    public static final DateTimeFormatter RFC3339_DATE_FORMAT = new DateTimeFormatterBuilder()	
        .appendPattern("yyyy-MM-dd'T'HH:mm:ss")	
        .appendFraction(ChronoField.NANO_OF_SECOND, 0, 9, true)	
        .appendOffsetId()	
        .toFormatter();
```
would work better than `java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME`

the above was used in previous version https://github.com/cloudevents/sdk-java/pull/216/files#diff-bac8dca098ba932b36f5b17870212371L34.

Or maybe we could change assertions in the test, I think that would be acceptable too.

/cc @slinkydeveloper

zhxnlai:
@matejvasek fixed the test. It looks like `ISO_OFFSET_DATE_TIME` actually removes unnecessary zeros in the test 😉

slinkydeveloper:
I personally don't care about the more zeroes, if that's compliant to the spec, fine. I would love to avoid adding custom formats

slinkydeveloper (approved):
@matejvasek can you double check this one?

matejvasek:
@zhxnlai my bad, I mistaken `expected` with `actual` in the test output.

matejvasek:
LGTM

matejvasek:
Note: when reading input we allow missing seconds, but for input it's OK to be less strict as is doesn't affect compatibility.

zhxnlai:
@matejvasek @slinkydeveloper any chance we could roll this out soon?

matejvasek:
@zhxnlai I am not maintainer here (@slinkydeveloper  is), but I would say probably not. This was merged just one day after releasing `2.0.0 Milestone 3`, and there hasn't been much done for [M4](https://github.com/cloudevents/sdk-java/milestone/4).

zhxnlai:
Technically this is a bug fix. I am waiting for this fix to upgrade my java application to the latest release so that it can interop with a go application

slinkydeveloper:
@zhxnlai hopefully another release will be out soon, we're trying to solve this issue first #250  :smile:

slinkydeveloper:
@zhxnlai Meanwhile you can use the snapshot packages, which should include this fix

Issue #238, opened 2020-09-25
CloudEvents with Avro payloads, on Kafka


mnuttall:
My team is looking at sending Avro-encoded cloud events over Kafka, often to or from Apache Camel-K. We’ve done some prototyping in which we’ve had to combine the Cloud Events Java SDK with the io.confluent.kafka Serde library. This leaves message receivers having to choose between the CloudEvent and Confluent deserializers since the former unpacks the headers nicely but produces an encoded payload, and the latter decodes the payload but leaves you to manage the headers yourself. 

This is different from the [Avro format for CloudEvents](https://github.com/cloudevents/spec/blob/v1.0/avro-format.md) which describes a CloudEvent in Avro - we want to do the opposite, sending a binary, Avro-encoded payload with headers as per the [Kafka Protocol Binding for CloudEvents](https://github.com/cloudevents/spec/blob/v1.0/kafka-protocol-binding.md). 

In [this Slack thread](https://cloud-native.slack.com/archives/CCXF3CBL1/p1601046799007500) I wondered, 

> (For example, one could pass `io.confluent.kafka.serializers.KafkaAvroDeserializer` to `io.cloudevents.kafka.CloudEventDeserializer` as something to delegate body-deserialization to, perhaps.) 

@slinkydeveloper stated, 
 
> That sounds like an interesting problem, there is no official answer from the sdk to the problem "body decoding". This was left out from the sdk design by purpose, in order to avoid making complex abstractions on top of marshallers/unmarshallers and/or opinionated choices. [...] Your solution of "delegation" for the kafka deserializer looks like an interesting choice. However, It solves the particular problem only for kafka: I would love to find a solution that can be applied in all the other reader/writers. 

He also mentioned [Issue 196](https://github.com/cloudevents/sdk-java/issues/196) as being somewhat related to this question. 

I would like to provide a nice way of handling messages with as much CloudEvents, schema management and de/serialization support built in as possible. Currently our prototype code has two layers: one using the CloudEvent APIs for handling the envelope then a second using Confluent APIs for handling the payload. Can anyone see a better way of doing this? Many thanks in advance for any suggestions!

slinkydeveloper:
I wonder if adding some "mapper" api like `CloudEvent#getData<T>(Function<Something, T>)` can at least ease the issue you're experiencing here.

slinkydeveloper:
This problem is kinda orthogonal to the one described in #198 but i see how we can find a solution to both problems at the same time

[...]

PR #250, opened 2020-10-07
Introduce CloudEventData

slinkydeveloper:
## Introduction

I thought a little bit about the problem with the data representation. To do a little recap, we have open these issues/PRs with various discussions about this topic:

* https://github.com/cloudevents/sdk-java/issues/196
* https://github.com/cloudevents/sdk-java/issues/238
* https://github.com/cloudevents/sdk-java/pull/198
* https://github.com/cloudevents/sdk-java/pull/208
* https://github.com/cloudevents/sdk-java/pull/211

I tried to put together all the different discussions we had, the different feedbacks and prepare a proposal to solve the issue, hopefully once and for all :smile: 

## What we're trying to solve

These are the problems we're trying to tackle:

* Users wants to read/write pojos in their CloudEvents, making the experience as nice as possible
* We want to enforce serialization/deserialization to bytes, but we don't want to couple to specific serde libraries
* We want an extensible representation
* We want to avoid double encoding/decoding while serializing/deserializing to any event format (look the #196 for an example outside the sdk), without coupling to a specific json library types

All these problems goes down to a problem: today data can be represented, by our interfaces, only with a byte array.

## How other sdks fix this issue

Before going deep into this proposal, I analyzed a bit what the other SDKs do:

* C# just serves `data` as `object` (the java `Object`). No particular handling is done about it
* Golang represents everything as byte array, so it suffers of the same issues we have
* Rust has a union type defined as string, bytes or json value that represents data. This is possible because in Rust there is one standard de-facto library to interact with json.

## How this new interface works

So this proposal is pretty simple: we add a layer of indirection between data and bytes with a new interface called `CloudEventData`. In particular:

* `getData()` now returns an implementation of `CloudEventData` (if any data). Whatever implementation is, it must be able to convert to bytes
* `CloudEventData` has a single method, `toBytes()`, that identify that going from `data` to bytes is a potentially expensive operation (because it may trigger a conversion from the internal type to bytes).
* 3rd party integrators can implement `CloudEventData` adding whatever they want: marshaller/unmarshaller to pojos, a specific type representing the integration with their own data type, ...

Then the change to `data` is ported to writer, making doable the optimizations to event format/protocol bindings. 

With this new interface, in order to support POJOs, in **core** we need to (in a followup, not this pr):

* Provide a subinterface like `POJOCloudEventData`
* Provide a concrete implementation of `POJOCloudEventData` (it may be `BaseCloudEvent` itself)
* `EventDataCodec` interface to marshal a pojo to bytes

This change doesn't modify the existing assumptions, nor relaxing or enforcing them. It just makes the existing interface more extensible.

What I think it's still a problem in this interface is that we force users to downcast often to retrieve the non bytes internal representation.

Note: This PR is incomplete, I would love to gather feedback first and then move forward with all the changes required to make it working.

cc @johanhaleby @aliok @pierDipi @bsideup 

### DIFF
--- a/api/src/main/java/io/cloudevents/CloudEvent.java
+++ b/api/src/main/java/io/cloudevents/CloudEvent.java
@@ -28,5 +28,5 @@ public interface CloudEvent extends CloudEventAttributes, CloudEventExtensions {
      * The event data
      */
     @Nullable
-    byte[] getData();
+    CloudEventData getData();
 }
--- a/api/src/main/java/io/cloudevents/CloudEventData.java
+++ b/api/src/main/java/io/cloudevents/CloudEventData.java
@@ -1,11 +1,7 @@
 package io.cloudevents;
 
-import java.util.function.Function;
-
 /**
- * Interface that defines a container for CloudEvent data.
- *
- * Every CloudEvent data has
+ * Interface that defines a wrapper for CloudEvent data.
  */
 public interface CloudEventData {
 
--- a/api/src/main/java/io/cloudevents/CloudEventData.java
+++ b/api/src/main/java/io/cloudevents/CloudEventData.java
@@ -2,10 +2,19 @@ package io.cloudevents;
 
 import java.util.function.Function;
 
+/**
+ * Interface that defines a container for CloudEvent data.
+ *
+ * Every CloudEvent data has
+ */
 public interface CloudEventData {
 
+    /**
+     * @return this CloudEventData, represented as bytes. Note: this operation may be expensive, depending on the internal representation of data
+     */
     byte[] toBytes();
 
+    //TBD
     <T> T to(Function<Object, T> mapper);
 
 }
--- a/api/src/main/java/io/cloudevents/CloudEventData.java
+++ b/api/src/main/java/io/cloudevents/CloudEventData.java
@@ -1,0 +1,11 @@
+package io.cloudevents;
+
+import java.util.function.Function;
+
+public interface CloudEventData {
+
+    byte[] toBytes();
+
+    <T> T to(Function<Object, T> mapper);
+
+}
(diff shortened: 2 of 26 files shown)


olegz:
Why do you insist on changing `CloudEvent` interface? 
This has already been extensively debated in #198, #208, #211 without you providing any constructive counter argument that all can agree and get behind.
I am somewhat speechless. . .

For the record it's a staunch -1 vote.

pierDipi (commented):
LGTM

olegz:
Additional changes did not address the core problem which introduces unnecessary type dependency in CloudEvent strategy.
Also, many of us stated in previous WIP PRs on the same subject, the core strategy (interface) must be simple and concise. Any kind of specialization, any kind of additional dependency could be introduced in extension strategies, providing the core framework can still accept the core interface. 

So definitive -1

JemDay:
If `CloudEventData` is the way I present the business data to the CloudEvent, how do I present a JsonValue or a ProtoMessage ? ... both JSON and Proto formats have the ability to carry the business data explicitly. 

Eg .. If I pass a [JsonValue](https://docs.oracle.com/javaee/7/api/javax/json/JsonValue.html) I'd hope/expect that the resultant CloudEvent in using a JSONFormat would have the JSON content I the `"data"` property.

Apologies if I'm missing something..

slinkydeveloper:
@JemDay that's the idea of this PR: with `CloudEventData` we allow carrying `JsonValue` or `ProtoMessage` as data, so with a downcast to `ProtoCloudEventData` or `JsonCloudEventData` you can actually get the "concrete data" you want. How we implement that in serializers/deserializers is still TBD

olegz:
@n3wscott @duglin @slinkydeveloper I am trying to understand how this change was actually merged, given that following the current and established community process this has been rejected several times providing detailed justification to which there was no followups?

PR #251, opened 2020-10-29
Introduce JsonCloudEventData

slinkydeveloper:
Followup of #250

This implements an optimization in the jackson CloudEvent ser/de that allows users to directly provide a `JsonNode` as `CloudEvent` data, avoiding the double serialization/deserialization problem.

This is inspired by the discussions here: https://github.com/cloudevents/sdk-java/issues/196

### DIFF
--- a/formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonCloudEventData.java
+++ b/formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonCloudEventData.java
@@ -1,0 +1,47 @@
+package io.cloudevents.jackson;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.cloudevents.CloudEventData;
+
+import java.util.Objects;
+
+/**
+ * This class is a wrapper for Jackson {@link JsonNode} implementing the {@link CloudEventData}
+ */
+public class JsonCloudEventData implements CloudEventData {
+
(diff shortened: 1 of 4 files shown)


formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonCloudEventData.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private final JsonNode node;
+
+    public JsonCloudEventData(JsonNode node) {
+        this.node = node;
+    }
+
+    @Override
+    public byte[] toBytes() {
+        return node.toString().getBytes();
```
pierDipi, line 21: Can you check if the node is null?

slinkydeveloper: It cannot be: if the `CloudEventData` is there, then there is something inside it. I guess I should add the check to the constructor of `JsonCloudEventData`

pierDipi (approved):

formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonCloudEventData.java
pierDipi: Yes, that is what I meant.

PR #257, opened 2020-11-10
CloudEventUtils.mapData(event, mapper)

slinkydeveloper:
This PR adds a new method to `CloudEventUtils`, called `mapData(event, mapper)` (implemented by default), to improve the user experience with mappers. Now we finally have a good user experience for data: users can now safely map event payloads to business objects without casting and without any knowledge of the sdk itself:

```java
MyPojo myPojo = mapData(event, PojoCloudEventDataMapper.from(objectMapper, new TypeReference<MyPojo>() {})) // PojoCloudEventDataMapper<MyPojo>
  .getValue();
```

(note: this example is using json mapper which is proposed in another pr independent from this one)

To make this possible, I also generified the return value of CloudEventDataMapper

### DIFF
--- a/core/src/main/java/io/cloudevents/core/message/MessageReader.java
+++ b/core/src/main/java/io/cloudevents/core/message/MessageReader.java
@@ -18,6 +18,7 @@
 package io.cloudevents.core.message;
 
 import io.cloudevents.CloudEvent;
+import io.cloudevents.CloudEventData;
 import io.cloudevents.core.builder.CloudEventBuilder;
 import io.cloudevents.lang.Nullable;
 import io.cloudevents.rw.*;
@@ -116,7 +117,7 @@ public interface MessageReader extends StructuredMessageReader, CloudEventReader
      * @throws CloudEventRWException if something went wrong during the visit.
      * @throws IllegalStateException if the message has an unknown encoding.
      */
-    default CloudEvent toEvent(@Nullable CloudEventDataMapper<?> mapper) throws CloudEventRWException, IllegalStateException {
+    default CloudEvent toEvent(@Nullable CloudEventDataMapper<? extends CloudEventData> mapper) throws CloudEventRWException, IllegalStateException {
         switch (getEncoding()) {
             case BINARY:
                 return this.read(CloudEventBuilder::fromSpecVersion, mapper);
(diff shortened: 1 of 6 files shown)


olegz:
As it was debated many times before this change violates the consistency and simplicity of the core interface.
So -1

Tagging @n3wscott @duglin as I already have an open complaint about a PR being merged with no consensus reached - https://github.com/cloudevents/sdk-java/pull/250#issuecomment-723984446

slinkydeveloper:
Related to https://github.com/cloudevents/sdk-java/pull/258

johanhaleby:
My feeling is that serialization/conversion should not be a part of the interface, it's a separate concern. 

It would be OK if it was an [extension function](https://kotlinlang.org/docs/reference/extensions.html) provided by the serialization library/module.

slinkydeveloper:
@johanhaleby Except we don't have extensions in java :smile: How about moving to some utils class, like `CloudEventUtils` in core?

slinkydeveloper:
Moved, now it should be better

core/src/main/java/io/cloudevents/core/message/MessageReader.java
```diff
@@ -116,7 +116,7 @@ default CloudEvent toEvent() throws CloudEventRWException, IllegalStateException
      * @throws CloudEventRWException if something went wrong during the visit.
      * @throws IllegalStateException if the message has an unknown encoding.
      */
-    default CloudEvent toEvent(@Nullable CloudEventDataMapper mapper) throws CloudEventRWException, IllegalStateException {
+    default CloudEvent toEvent(@Nullable CloudEventDataMapper<?> mapper) throws CloudEventRWException, IllegalStateException {
```
aliok, line 119: would a `CloudEventDataMapper<? extends CloudEventData>` work? Haven't tried with a compiler though

core/src/main/java/io/cloudevents/core/message/StructuredMessageReader.java
```diff
@@ -44,7 +44,7 @@ default CloudEvent toEvent() throws CloudEventRWException, IllegalStateException
         return this.read(EventFormat::deserialize);
     }
 
-    default CloudEvent toEvent(@Nullable CloudEventDataMapper mapper) throws CloudEventRWException, IllegalStateException {
+    default CloudEvent toEvent(@Nullable CloudEventDataMapper<?> mapper) throws CloudEventRWException, IllegalStateException {
```
aliok, line 47: same here

aliok:
This seems like the next natural step after the previous improvements.
Makes more sense with https://github.com/cloudevents/sdk-java/pull/258.

/lgtm

pierDipi (approved):
LGTM

PR #267, opened 2020-11-12
Javadoc'ed + Cleanup of the api module

slinkydeveloper:
Also:

* Cleanup the `CloudEventRWException`
* Added a bunch of tests in api module
* Added in `Time` overrides of `parseTime` and `writeTime` to generate the proper `CloudEventRWException`. Also using this methods when appropriate

### DIFF
--- a/api/src/main/java/io/cloudevents/types/Time.java
+++ b/api/src/main/java/io/cloudevents/types/Time.java
@@ -17,6 +17,9 @@
 
 package io.cloudevents.types;
 
+import io.cloudevents.rw.CloudEventRWException;
+
+import java.time.DateTimeException;
 import java.time.OffsetDateTime;
 import java.time.format.DateTimeParseException;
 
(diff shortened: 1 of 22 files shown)


matejvasek (approved):

slinkydeveloper:
@matejvasek can you review this again? I rebased

api/src/main/java/io/cloudevents/types/Time.java
```diff
@@ -31,16 +34,56 @@ private Time() {
...
+
+    /**
+     * Convert an attribute/extension {@link OffsetDateTime} to a RFC3339 compliant {@link String}.
+     *
+     * @param attributeName the attribute/extension name
+     * @param time          the time to write as {@link String}
+     * @return the serialized time
+     * @throws CloudEventRWException if something went wrong when serializing the attribute/extension.
+     */
+    public static String writeTime(String attributeName, OffsetDateTime time) throws DateTimeException {
```
matejvasek, line 82: When I first saw call to this method I was confused why it needs attributeName. I found unusual that `writeTime()` needs that. Maybe keep explicit exception catching?

slinkydeveloper: So you propose to remove these overrides?

matejvasek: At first I thought we should remove the overload because the overload is used only on one place. But now I see that that there is similar `parseTime()` method which is used at multiple places. So better keep it as it is.

matejvasek (approved):

Issue #30, opened 2019-01-07
AMQP 1.0 transport


matzew:
We should start evaluating to add AMQP 1.0 support, e.g. using Apache QPid's proton-j lib.

Alfusainey:
@slinkydeveloper I want to work on this issue. can I give it a go?

slinkydeveloper:
Sure, go ahead. Which amqp client are you planning to use? Do you want to use apache qpid jproton?

Alfusainey:
I am planning to use the [vertx-proton](https://github.com/vert-x3/vertx-proton) library to integrate  amqp 1.0 with vertx. This library is a wrapper around qpid proton-j

```XML
<dependency>
    <groupId>io.vertx</groupId>
    <artifactId>vertx-proton</artifactId>
    <version>${vertx.version}</version>
</dependency>
```

I see that different components in the project uses a different vertx version. Is there a reason for this?

How about defining a maven property for the vertx version (i.e `vertx.version`) in the parent POM and have all modules use the version defined in the parent? This way, the new `amqp` module will not have to redefine its own vertx version

slinkydeveloper:
IMO we should integrate directly with jproton more than the vertx wrapper, because it might be useful for folks that needs amqp but doesn't want to use vertx.

> I see that different components in the project uses a different vertx version. Is there a reason for this?

Which components?

Alfusainey:
> IMO we should integrate directly with jproton more than the vertx wrapper, because it might be useful for folks that needs amqp but doesn't want to use vertx

ok 👍 

> Which components?

http/vertx uses `4.0.0.Beta1` while http/restful-ws uses `3.9.0

slinkydeveloper:
> http/vertx uses 4.0.0.Beta1 while http/restful-ws uses 3.9.0

No issue then

Alfusainey:
@slinkydeveloper If i understand correctly, the amqp protocol binding spec does not mandate the AMQP section to store CE extensions.

> CloudEvents extensions that define their own attributes MAY define a secondary mapping to AMQP properties for those attributes, also in different message sections, especially if specific attributes or their names need to align with AMQP features or with other specifications that have explicit AMQP header bindings

So the mapping can go either to the properties section of the message or to a different section chosen by the user. I am thinking of implementing this as follows:

1. Map all extensions to the application-properties section of the message. In this case, all attributes and extensions maps to the same section of the message. I think this is what the sdk-go impl. is doing
[...]

Issue #264, opened 2020-11-11
Documentation for 2.0


slinkydeveloper:
This is an umbrella issue to cover all the things we need regarding documentation for 2.0:

* [x] Examples: #125
* [x] Enable javadoc failures when running `mvn verify` (mvn javadoc plugin) #275
* [x] Check missing javadoc, missing javadoc tags and internal and external links are working properly
    * [x] api #267 
    * [x] core https://github.com/cloudevents/sdk-java/pull/313
    * [x] formats/json-jackson https://github.com/cloudevents/sdk-java/pull/319
    * [x] http/basic #320
    * [x] http/restful-ws #320
    * [x] http/vertx #320
    * [x] kafka #321
    * [x] amqp #322
    * [x] spring #323
* [ ] Update website:
  * [x] Check website technical details (jekyll version, theme version, ...)
  * [x] Develop a page that shows the basic features to operate with CloudEvents: CloudEvent, CloudEventBuilder, how to interact with `CloudEventData` and map it to pojos, ... #324
  * [x] Develop a page that shows the basics of using `EventFormat`, showing an example for json (using jackson) #324
  * [x] Move readmes of separate modules to website pages, one page per module (and point modules back to website pages) #324
    * [x] core and api (they might require a single page for both) #324
    * [x] formats/json-jackson #324
    * [x] http/basic #324
    * [x] http/restful-ws #324
    * [x] http/vertx #324
    * [x] kafka #324
  * [ ] Develop a page that explains some advanced stuff, like how to create your own `CloudEvent`, how to create a protocol binding, how to create your own event format, etc

If you're going to work in one of the issues described above, please write it down in the comments.

slinkydeveloper:
I'm gonna start working on the technical details of the website

slinkydeveloper:
Checked javadocs for api module: https://github.com/cloudevents/sdk-java/pull/267

slinkydeveloper:
Run javadoc builds with mvn verify https://github.com/cloudevents/sdk-java/pull/275

slinkydeveloper:
Working on core module docs

slinkydeveloper:
Working on jackson

slinkydeveloper:
Working on all the http/* modules javadocs

slinkydeveloper:
Working on kafka javadocs

slinkydeveloper:
Working on amqp javadocs

slinkydeveloper:
Working on spring javadocs

[...]

PR #270, opened 2020-11-13
[#30] Implement AMQP 1.0 transport binding

Alfusainey:
An initial patch implementing the AMQP 1.0 protocol binding for client events. This patch implements the `MessageReader` and `MessageWriter` interfaces to support the reading and writing of cloud events to and from an AMQP representation

@slinkydeveloper I did a demo integration test using a vertx-proton client and server to send and receive cloud event messages. once this is merged, i will provide a sample README

Signed-off-by: Alfusainey Jallow <alf.jallow@gmail.com>


amqp/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+
+	<artifactId>cloudevents-amqp</artifactId>
+	<name>CloudEvents - AMQP Binding</name>
+	<packaging>jar</packaging>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.apache.qpid</groupId>
+			<artifactId>proton-j</artifactId>
+			<version>0.33.7</version>
```
slinkydeveloper, line 20: Make this version a property

amqp/src/main/java/io/cloudevents/amqp/AmqpBinaryMessageReader.java
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright 2020-Present The CloudEvents Authors
```
slinkydeveloper, line 2: AFAIR, this should be 2018-Present, can you check the others too?

amqp/src/main/java/io/cloudevents/amqp/AmqpBinaryMessageReader.java
```diff
@@ -0,0 +1,127 @@
...
+     * This method simply returns the string representation of the type of value passed as argument.
+     * 
+     * @param value The value of a CloudEvent attribute or extension.
+     * @return The string representation of the specified value.
+     * 
+     * @throws NullpointerException if the value is {@code null}.
+     */
+    @Override
+    protected String toCloudEventsValue(final Object value) {
+        return Objects.requireNonNull(value).toString();
```
slinkydeveloper, line 125: Can the attribute be there in the applicationProperties map but with a null value?

amqp/src/main/java/io/cloudevents/amqp/AmqpMessageFactory.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.qpid.proton.message.Message;
+
+import io.cloudevents.core.message.MessageReader;
+import io.cloudevents.core.message.impl.GenericStructuredMessageReader;
+import io.cloudevents.core.message.impl.MessageUtils;
+import io.cloudevents.core.message.impl.UnknownEncodingMessageReader;
+
+public class AmqpMessageFactory {
+
+    public static MessageReader createReader(final Message message) {
```
slinkydeveloper, line 35: All the public static methods here should be well documented, these are the entrypoints of the module. Also, if you can, add something to root javadoc of the class

amqp/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>io.cloudevents</groupId>
+		<artifactId>cloudevents-parent</artifactId>
+		<version>2.0.0-SNAPSHOT</version>
+	</parent>
+
+	<artifactId>cloudevents-amqp</artifactId>
```
slinkydeveloper, line 12: should we name it following the same pattern of other "integration modules" aka `cloudevents-[protocol]-[library]`? Or better, do we expect someday someone comes in proposing a new amqp integration with another library (eg http integrations)? Or is it proton-j a standard de facto client, pretty much like official java kafka client is the standard de facto client for kafka?

amqp/src/main/java/io/cloudevents/amqp/AmqpBinaryMessageReader.java
```diff
@@ -0,0 +1,127 @@
+/*
```
slinkydeveloper, line 1: Can we have this class under a subpackage `impl`?

amqp/src/main/java/io/cloudevents/amqp/AmqpBinaryMessageWriter.java
```diff
@@ -0,0 +1,104 @@
+/*
```
slinkydeveloper, line 1: Can we have this class under a subpackage `impl`?

amqp/src/main/java/io/cloudevents/amqp/AmqpConstants.java
```diff
@@ -0,0 +1,51 @@
+/*
```
slinkydeveloper, line 1: Can we have this class under a subpackage `impl`?

amqp/src/main/java/io/cloudevents/amqp/AmqpMessageFactory.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.qpid.proton.amqp.messaging.Data;
+import org.apache.qpid.proton.amqp.messaging.Section;
+import org.apache.qpid.proton.message.Message;
+
+import io.cloudevents.core.message.MessageReader;
+import io.cloudevents.core.message.impl.GenericStructuredMessageReader;
+import io.cloudevents.core.message.impl.MessageUtils;
+import io.cloudevents.core.message.impl.UnknownEncodingMessageReader;
+
+public class AmqpMessageFactory {
```
slinkydeveloper, line 33: Make it public final and add private constructor, also annotate with @ParametersAreNonnullByDefault

amqp/src/main/java/io/cloudevents/amqp/AmqpMessageFactory.java
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @param <T> The expected type of the property to retrieve.
+     * @param props The application properties to retrieve the value from.
+     * @param name The name of the application property.
+     * @param type The expected value type.
+     * @return The value or {@code null} if the properties do not contain a value of the expected type for the given
+     *         name.
+     */
+    @SuppressWarnings("unchecked")
+    private static <T> T getApplicationProperty(final ApplicationProperties props, final String name,
```
slinkydeveloper, line 68: Can we have these methods somewhere else like in a class `AmqpUtils` under the subpackage `impl`?

amqp/src/main/java/io/cloudevents/amqp/AmqpBinaryMessageWriter.java
```diff
@@ -0,0 +1,104 @@
...
+                throw new IllegalStateException("This Writer is not initialized");
+            }
+            applicationProperties.getValue().put(AmqpConstants.ATTRIBUTES_TO_PROPERTYNAMES.get(name), value);
+        }
+        return null;
+    }
+
+    @Override
+    public CloudEventExtensionsWriter withExtension(String name, String value) throws CloudEventRWException {
+        // TODO: for now, extensions are mapped to application-properties
```
slinkydeveloper, line 67: WDYM? Do we expect to map them somewhere else?

amqp/src/main/java/io/cloudevents/amqp/AmqpBinaryMessageWriter.java
```diff
@@ -0,0 +1,104 @@
...
+            applicationProperties = new ApplicationProperties(new HashMap<String, Object>());
+        }
+        applicationProperties.getValue().put(AmqpConstants.APP_PROPERTY_SPEC_VERSION, version.toString());
+        return this;
+    }
+
+    @Override
+    public Message setEvent(EventFormat format, byte[] value) throws CloudEventRWException {
+        message.setContentType(format.serializedContentType());
+        return this.end(new BytesCloudEventData(value));
```
slinkydeveloper, line 87: Can you duplicate the code of end here to avoid this wrapping?

amqp/src/main/java/io/cloudevents/amqp/AmqpMessageFactory.java
```diff
@@ -0,0 +1,103 @@
...
+            final AmqpValue body = (AmqpValue) payload;
+            if (body.getValue() instanceof byte[]) {
+                return (byte[]) body.getValue();
+            } else if (body.getValue() instanceof String) {
+                return ((String) body.getValue()).getBytes(StandardCharsets.UTF_8);
+            }
+        }
+
+        return null;
+    }
```
slinkydeveloper, line 101: After we merge this one, it might be interesting to investigate if we can leverage the `CloudEventData` abstraction to perform some optimizations here (eg when you receive an event from amqp, change something in the attributes, and then send it back to amqp)

slinkydeveloper: Pretty much like we did for json https://github.com/cloudevents/sdk-java/pull/251

slinkydeveloper (commented):
This PR looks reaaaally nice and I'm super excited about it, thanks for your contribution!!!

I left a bunch of questions and style comments. Sorry in advance for the comments related to my poor knowledge of amqp :smile:

matzew:
REally nice work! Glad to see this coming!

amqp/src/main/java/io/cloudevents/amqp/AmqpBinaryMessageReader.java
Alfusainey: ok, will change accordingly.. i was thinking the start date indicates the date file was created.

amqp/src/main/java/io/cloudevents/amqp/AmqpBinaryMessageReader.java
Alfusainey: i think this depends on how strict the map implementation used for the applicationProperties is. This reader uses the available HashMap implementation which is less strict and thus allows `null` values (and even null keys). since the `ApplicationProperties` is defined to take any Map type, i added this just to be on the safe side

but i can check if there is already a stricter map implementation for app properties present within proton-j

amqp/src/main/java/io/cloudevents/amqp/AmqpMessageFactory.java
Alfusainey: sure 👍

amqp/pom.xml
Alfusainey: good question! i dont know the answer TBH. 

i would think of qpid-proton more as a library that can be integrated in any AMQP client/broker/etc. One can use a [`JMS`](https://www.oracle.com/technical-resources/articles/java/intro-java-message-service.html) client, or the rabbit-mq client and there might be other clients too. these clients may not rely on proton-j though.

so i think we can use `cloudevents-[protocol]-[library]`, WDYT?

amqp/src/main/java/io/cloudevents/amqp/AmqpMessageFactory.java
Alfusainey: how about putting them in `AmqpConstants.java`?

amqp/src/main/java/io/cloudevents/amqp/AmqpBinaryMessageWriter.java
Alfusainey: this was the [comment](https://github.com/cloudevents/sdk-java/issues/30#issuecomment-723982190) i raised in the issue. 

the spec does not mandate the location where extensions can be stored (and left this to the implementors). your suggestion, as a first pass, was to put them in the application-properties section of the message. but it could be that the extension names may align with the name of amqp message properties and a user of the library would like to store the extensions with the message properties because they align better.

this is why i added this `TODO` note here

amqp/src/main/java/io/cloudevents/amqp/AmqpMessageFactory.java
Alfusainey: yes, this makes a lot of sense

amqp/src/main/java/io/cloudevents/amqp/AmqpBinaryMessageReader.java
slinkydeveloper: It was agreed in past that we should use the same for all the files since the beginning of the project

amqp/src/main/java/io/cloudevents/amqp/AmqpMessageFactory.java
slinkydeveloper: That's fine too

amqp/pom.xml
slinkydeveloper: yeah let's go with `cloudevents-amqp-proton` ok?

amqp/src/main/java/io/cloudevents/amqp/AmqpBinaryMessageReader.java
slinkydeveloper: I think it's fine, my only concern is that, if the value is null, then this method shouldn't be called at all. Maybe you need the null check in `forEachHeader`?

amqp/src/main/java/io/cloudevents/amqp/AmqpBinaryMessageWriter.java
slinkydeveloper: I think you can even remove the todo and just leave the comment (even better, add the link to the GH comment where you talked about this)

amqp/pom.xml
Alfusainey: sure, should i then change the name of the class to something with `proton` in the name? e.g `ProtonBasedAmqpBinaryMessageWriter`?

Alfusainey:
i updated the PR to incorporate the comments. i also renamed the class names to include `Proton` so that it is clear which amqp library the implementation uses

amqp/src/main/java/io/cloudevents/amqp/AmqpMessageFactory.java
Alfusainey: @slinkydeveloper created #276 to track this

amqp/src/main/java/io/cloudevents/amqp/ProtonBasedAmqpMessageFactory.java
```diff
@@ -0,0 +1,106 @@
...
+import io.cloudevents.core.message.impl.GenericStructuredMessageReader;
+import io.cloudevents.core.message.impl.MessageUtils;
+import io.cloudevents.core.message.impl.UnknownEncodingMessageReader;
+import io.cloudevents.rw.CloudEventWriter;
+/**
+ * A factory class providing convenience methods for creating MessageReader and MessageWriter instances based on
+ * {@link <a href="https://github.com/apache/qpid-proton-j">Qpid Proton</a>}.
+ */
+@ParametersAreNonnullByDefault
+public final class ProtonBasedAmqpMessageFactory {
```
slinkydeveloper, line 40: I think without Based is fine, let's not be too much verbose :smile: Same for other classes

amqp/src/test/java/io/cloudevents/amqp/ProtonBasedAmqpMessageWriterTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+        final MessageWriter<?, Message> writer = new ProtonBasedAmqpMessageWriter<Message>();
+        final Message actualMessage = writer.writeStructured(event, format.serializedContentType());
+
+        assertThat(actualMessage.getContentType()).isEqualTo(expectedMessage.getContentType());
+        assertThat(actualMessage.getBody().toString()).isEqualTo(expectedMessage.getBody().toString());
+        assertThat(actualMessage.getApplicationProperties()).isNull();
+    }
+
+    private Message translateBinaryEvent(final CloudEvent event) {
```
slinkydeveloper, line 83: Can you try a message with cloudevents extensions too?

slinkydeveloper (commented):
Small nits and we can merge

amqp/src/main/java/io/cloudevents/amqp/ProtonBasedAmqpMessageFactory.java
Alfusainey: sure 😄

amqp/src/test/java/io/cloudevents/amqp/ProtonBasedAmqpMessageWriterTest.java
Alfusainey: i notice a small bug(?) while testing this. 

In `io.cloudevents.SpecVersion`, the `datacontentencoding` attribute name is included in the list of optional attributes for the V03 (i.e in CloudEventAttributes#getAttributeNames()). However, when i use that name to retrieve its value using CloudEventV03#getAttribute(String), i get an `IllegalArgumentException:  Spec version v0.3 doesn't have attribute named datacontentencoding`;

i think an additional branch should be added to the switch statement to handle the case when the attributeName is `datacontentencoding`. Since the value of this attribute is not stored inside a CloudEventV03 object, we can just do nothing in this case, instead of raising an exception. any thoughts, @slinkydeveloper ? this is what am doing now to make my tests happy 😄

slinkydeveloper: The `datacontentencoding` is a funny one :smile: In the 0.3 spec, it was a bit unclear its role outside the json itself. So in this sdk, I decided to don't save it in the event, because it doesn't make sense to do so. It is in fact only used in the json format, when the event is marshalled/unmarshalled back and forth the wire.

I guess the right behaviour here is that, when `CloudEventV03#getAttribute(String)` is invoked with `datacontentencoding`, it should just return null always.

slinkydeveloper: I can provide a patch for it, one moment

slinkydeveloper: Here it is https://github.com/cloudevents/sdk-java/pull/279 Mind doing a review?

Alfusainey: ok, done

slinkydeveloper:
@Alfusainey I'm looking now at the errors and it seems they're related to this pr, let me show you

amqp/src/main/java/io/cloudevents/amqp/impl/ProtonAmqpBinaryMessageReader.java
```diff
@@ -0,0 +1,131 @@
...
+public final class ProtonAmqpBinaryMessageReader extends BaseGenericBinaryMessageReaderImpl<String, Object> {
+
+    private final String contentType;
+    private final ApplicationProperties applicationProperties;
+
+    /**
+     * Create an instance of an AMQP message reader.
+     *
+     * @param version                  The version of the cloud event message.
+     * @param applicationPropereties   The application properties of the AMQP message that contains
```
slinkydeveloper, line 51: Typo in applicationProperties

amqp/src/main/java/io/cloudevents/amqp/impl/ProtonAmqpBinaryMessageReader.java
```diff
@@ -0,0 +1,131 @@
...
+     *                                 the cloud event metadata (i.e attributes and extensions).
+     *                                 The applicationProperties MUST not be {@code null}.
+     * @param contentType              The content-type property of the AMQP message or {@null} if the message content type is unknown.
```
slinkydeveloper, line 54: s/{@null}/{@code null}

amqp/src/main/java/io/cloudevents/amqp/ProtonAmqpMessageFactory.java
```diff
@@ -0,0 +1,106 @@
...
+import io.cloudevents.amqp.impl.AmqpConstants;
+import io.cloudevents.core.message.MessageReader;
+import io.cloudevents.core.message.MessageWriter;
+import io.cloudevents.core.message.impl.GenericStructuredMessageReader;
+import io.cloudevents.core.message.impl.MessageUtils;
+import io.cloudevents.core.message.impl.UnknownEncodingMessageReader;
+import io.cloudevents.rw.CloudEventWriter;
+/**
+ * A factory class providing convenience methods for creating MessageReader and MessageWriter instances based on
+ * {@link <a href="https://github.com/apache/qpid-proton-j">Qpid Proton</a>}.
```
slinkydeveloper, line 37: You don't need to wrap this in a link

amqp/src/main/java/io/cloudevents/amqp/impl/ProtonAmqpBinaryMessageReader.java
Alfusainey: thanks for catching 👍

Alfusainey:
> @Alfusainey I'm looking now at the errors and it seems they're related to this pr, let me show you

@slinkydeveloper the build is actually failing to generate javadoc for the cloudevents-api module. here is a summary:

```
[ERROR] Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in http://docs.oracle.com/javase/8/docs/api/ are in the unnamed module.
[ERROR] sdk-java/api/src/main/java/io/cloudevents/CloudEvent.java:31: warning: no @return
[ERROR]     CloudEventData getData();
[ERROR]                    ^
[ERROR] sdk-java/api/src/main/java/io/cloudevents/rw/CloudEventReader.java:38: warning: no @param for <V>
[ERROR]     default <V extends CloudEventWriter<R>, R> R read(CloudEventWriterFactory<V, R> writerFactory) throws CloudEventRWException {
[ERROR]                                                  ^
[ERROR] sdk-java/api/src/main/java/io/cloudevents/rw/CloudEventReader.java:38: warning: no @param for <R>
[ERROR]     default <V extends CloudEventWriter<R>, R> R read(CloudEventWriterFactory<V, R> writerFactory) throws CloudEventRWException {
[ERROR]                                                  ^
[ERROR] sdk-java/api/src/main/java/io/cloudevents/rw/CloudEventReader.java:38: warning: no @return
[ERROR]     default <V extends CloudEventWriter<R>, R> R read(CloudEventWriterFactory<V, R> writerFactory) throws CloudEventRWException {
[ERROR]                                                  ^
[ERROR] sdk-java/api/src/main/java/io/cloudevents/rw/CloudEventReader.java:45: warning: no @param for <V>
[ERROR]     <V extends CloudEventWriter<R>, R> R read(CloudEventWriterFactory<V, R> writerFactory, @Nullable CloudEventDataMapper mapper) throws CloudEventRWException;
[ERROR]                                          ^
[ERROR] sdk-java/api/src/main/java/io/cloudevents/rw/CloudEventReader.java:45: warning: no @param for <R>
[ERROR]     <V extends CloudEventWriter<R>, R> R read(CloudEventWriterFactory<V, R> writerFactory, @Nullable CloudEventDataMapper mapper) throws CloudEventRWException;
```
any idea what am doing wrong?

amqp/src/test/java/io/cloudevents/amqp/ProtonAmqpMessageWriterTest.java
```diff
@@ -21,43 +21,47 @@
...
-        final Message expectedMessage = translateBinaryEvent(binaryEvent);
+    @ParameterizedTest()
+    @MethodSource("io.cloudevents.core.test.Data#allEventsWithStringExtensions")
+    public void testWriteBinaryCloudEventToAmqpRepresentation(final CloudEvent binaryEvent) {
+
+        Message expectedMessage = null;
+        try {
+            expectedMessage = translateBinaryEvent(binaryEvent);
+        } catch (final Exception e) {
+            System.out.println(e);
```
Alfusainey, line 56: i used this for debugging but forgot to remove. will include this in the final update

Alfusainey:
i needed to modify the maven-javadoc-plugin configuration for the build to work for me (https://stackoverflow.com/a/61884267/8462878). i will not commit this change though :-)

slinkydeveloper:
> @slinkydeveloper the build is actually failing to generate javadoc for the cloudevents-api module

So on your machine javadoc tool fails for warnings? Interesting... Anyway, we plan to fix javadoc properly soon https://github.com/cloudevents/sdk-java/issues/264 so hopefully this will be fixed asap.

slinkydeveloper:
@Alfusainey thanks again for this contribution :rocket:

Alfusainey:
> @Alfusainey thanks again for this contribution 🚀

you're welcome, and thanks a lot for taking the time to review it! I will be hanging around this project from now on 😄

slinkydeveloper:
Oh yeah, feel free to check out the other good first issues https://github.com/cloudevents/sdk-java/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22 and also join the cncf slack if you want to https://cloud-native.slack.com/archives/CCXF3CBL1

amqp/src/main/java/io/cloudevents/amqp/ProtonAmqpMessageFactory.java
```diff
@@ -0,0 +1,105 @@
...
+     * Creates a MessageReader using the content-type property and payload of a proton-based message.
+     * <p>
+     * This method simply calls {@link #createReader(String, ApplicationProperties, byte[])}.
+     * 
+     * @param contentType The content-type of the message payload.
+     * @param payload     The message payload in bytes.
+     * @return            A message reader capable of representing a CloudEvent from
+     *                    a message <em>content-type</em> property and <em>application-data</em>.
+     */
+    public static MessageReader createReader(final String contentType, final byte[] payload) {
```
slinkydeveloper, line 70: Hi @Alfusainey I just noticed this method and i was wondering... do we really need it? Can you provide a use case for it?

Issue #262, opened 2020-11-11
Remove UnknownEncodingMessageReader and replace with exceptions


slinkydeveloper:
* Remove `UnknownEncodingMessageReader` and, in `MessageUtils#parseStructuredOrBinaryMessage`, replace the unknownMessageFactory with a `CloudEventRWException`
* Removed `Encoding#UNKNOWN` variant since it won't be useful anymore

smadasu:
@slinkydeveloper TBH, I don't even see the need for this class at all. There is no functionality in the class except throwing exceptions for any method call. This class should be removed and CloudEventRWException should be thrown at the places where this is instantiated.

slinkydeveloper:
Makes sense, do you wish to contribute it @smadasu?

smadasu:
Sure @slinkydeveloper  . I will create a PR for this.

slinkydeveloper:
Solved by #277

PR #277, opened 2020-11-15
Remove UnknownEncodingMessageReader and replace with exceptions #262

smadasu:
Removed UNKNOWN encoding and replaced class with Exception. Please approve.


slinkydeveloper:
Can you fix the dco? If you look in the details, it shows you how

core/src/main/java/io/cloudevents/core/message/impl/MessageUtils.java
```diff
@@ -57,7 +59,7 @@ public static MessageReader parseStructuredOrBinaryMessage(
             return binaryMessageFactory.apply(SpecVersion.parse(specVersionUnparsed));
         }
 
-        return unknownMessageFactory.get();
+        throw CloudEventRWException.newOther(new IllegalStateException("Invalid content type or spec version"));
```
slinkydeveloper, line 62: There is a CloudEventRWException kind for it right? If not, can we create it?

slinkydeveloper (changes requested):
Can you add a test for `parseStructuredOrBinaryMessage` that tests all three cases (encoding binary, encoding structured and exception thrown for unknown encoding)?

core/src/main/java/io/cloudevents/core/message/impl/MessageUtils.java
smadasu: I will fix the dco. As for that exception, this call is creating that exception

smadasu: I will work on the test

slinkydeveloper: I would love to have a specific `CloudEventRWException` for this, adding a kind here: https://github.com/cloudevents/sdk-java/blob/master/api/src/main/java/io/cloudevents/rw/CloudEventRWException.java#L28

api/src/main/java/io/cloudevents/rw/CloudEventRWException.java
```diff
@@ -56,6 +56,10 @@
          * Error while converting CloudEventData.
          */
         DATA_CONVERSION,
+        /**
+         * Invalid content type or spec version
+         */
+        PARSE_EXCEPTION,
```
slinkydeveloper, line 62: I think this should be named `UNKNOWN_ENCODING`, `PARSE_EXCEPTION` is too generic.

slinkydeveloper (changes requested):

api/src/main/java/io/cloudevents/rw/CloudEventRWException.java
smadasu: OK. Sure

smadasu: I have updated everything requested but the build is still failing for some reason. Please look into it.

slinkydeveloper:
SGTM, can you fix the broken tests?

slinkydeveloper:
Maybe you need a rebase on latest upstream/master

slinkydeveloper:
Thanks for the contribution, I solved the broken test!

PR #280, opened 2020-11-16
Removing ContextAttributes enums

slinkydeveloper:
They're useless and they're just increasing the size of the sdk


pierDipi:
We're actually using them in `eventing-kafka-broker`, so they're not so useless.

slinkydeveloper:
How are you using them? They're really necessary?

core/src/main/java/io/cloudevents/core/v03/CloudEventV03.java
```diff
@@ -36,6 +36,16 @@
...
 
+    private final static String ID = "id";
+    private final static String SOURCE = "source";
+    private final static String SPECVERSION = "specversion";
+    private final static String TYPE = "type";
+    private final static String TIME = "time";
+    private final static String SCHEMAURL = "schemaurl";
+    private final static String DATACONTENTTYPE = "datacontenttype";
+    private final static String DATACONTENTENCODING = "datacontentencoding";
+    private final static String SUBJECT = "subject";
```
pierDipi, line 47: Can you make those one public?

core/src/main/java/io/cloudevents/core/v1/CloudEventV1.java
```diff
@@ -35,6 +35,15 @@
...
 public final class CloudEventV1 extends BaseCloudEvent {
 
+    private final static String ID = "id";
+    private final static String SOURCE = "source";
+    private final static String SPECVERSION = "specversion";
+    private final static String TYPE = "type";
+    private final static String TIME = "time";
+    private final static String DATASCHEMA = "dataschema";
+    private final static String DATACONTENTTYPE = "datacontenttype";
+    private final static String SUBJECT = "subject";
```
pierDipi, line 45: Same here.

core/src/main/java/io/cloudevents/core/v03/CloudEventV03.java
slinkydeveloper: https://github.com/cloudevents/sdk-java/pull/282#issuecomment-729606022

pierDipi: We can hardcode them or if you make those ones public we can use them.

slinkydeveloper:
@Alfusainey @pierDipi wanna look again at this? I added javadocs + made constants public

pierDipi (approved):

PR #284, opened 2020-11-18
Add sources plugin to Maven build

dsyer:
It's nice to have sources in the IDE.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,6 +165,20 @@
                     </execution>
                 </executions>
             </plugin>
+            <!-- Sources jar -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-source-plugin</artifactId>
+                <version>3.2.1</version>
+                <executions>
+                    <execution>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -165,6 +165,20 @@
                     </execution>
                 </executions>
             </plugin>
+            <!-- Sources jar -->
+            <plugin>
```
slinkydeveloper, line 169: We already have this in the release profile, do we need it 2 times?

slinkydeveloper (commented):
Thanks for your contribution, I have just a small comment about this

pom.xml
dsyer: I didn't notice, sorry. Why would you want it in a profile though? If I just `mvn clean install` locally, I would want to see sources. Can I remove it from the profile?

slinkydeveloper: I think you can remove it from the profile. The release profile is used both for snapshots and production releases

slinkydeveloper: Ok as is now it's fine

PR #286, opened 2020-11-19
Introduce CloudEventDataMapper.identity()

### LINKED ISSUE #285: Don't allow null `CloudEventDataMapper`
I was starting to log an issue about a NullPointerException in Milestone 4 when I discovered that @matejvasek had already fixed it in #259. I'd like to suggest that other similar NPEs could be avoided if we didn't allow the mapper in [`MessageReader.toEvent(CloudEventDataMapper)`](https://www.javadoc.io/static/io.cloudevents/cloudevents-core/2.0.0-milestone4/io/cloudevents/core/message/MessageReader.html#toEvent-io.cloudevents.rw.CloudEventDataMapper-) to be null. There's no reason to allow that, and it's just setting a trap. If you don't need to map the data then just supply an identity mapper. The no-arg [`MessageReader.toEvent()`](https://www.javadoc.io/static/io.cloudevents/cloudevents-core/2.0.0-milestone4/io/cloudevents/core/message/MessageReader.html#toEvent--) could call `toEvent(data -> data)` instead of `toEvent(null)`. And a bunch of `if (mapper == null)` code could presumably be removed.

slinkydeveloper:
Fix #285

* Now, across all the different interfaces we have, `CloudEventDataMapper` cannot be null
* Cleanup of all mapper != null and replaced with CloudEventDataMapper.identity()
* Fixed all the missing wildcards

### DIFF
--- a/api/src/main/java/io/cloudevents/rw/CloudEventDataMapper.java
+++ b/api/src/main/java/io/cloudevents/rw/CloudEventDataMapper.java
@@ -40,7 +40,7 @@ public interface CloudEventDataMapper<R extends CloudEventData> {
     /**
      * No-op identity mapper which can be used as default when no mapper is provided.
      */
-    static CloudEventDataMapper<? extends CloudEventData> identity() {
+    static CloudEventDataMapper<CloudEventData> identity() {
         return d -> d;
     }
 }
--- a/api/src/main/java/io/cloudevents/rw/CloudEventDataMapper.java
+++ b/api/src/main/java/io/cloudevents/rw/CloudEventDataMapper.java
@@ -38,7 +38,9 @@ public interface CloudEventDataMapper<R extends CloudEventData> {
     R map(CloudEventData data) throws CloudEventRWException;
 
     /**
-     * No-op mapper which can be used as default when no mapper is provided
+     * No-op identity mapper which can be used as default when no mapper is provided.
      */
-    CloudEventDataMapper<?> NOOP = d -> d;
+    static CloudEventDataMapper<? extends CloudEventData> identity() {
+        return d -> d;
+    }
 }
--- a/api/src/main/java/io/cloudevents/rw/CloudEventDataMapper.java
+++ b/api/src/main/java/io/cloudevents/rw/CloudEventDataMapper.java
@@ -37,4 +37,8 @@ public interface CloudEventDataMapper<R extends CloudEventData> {
      */
     R map(CloudEventData data) throws CloudEventRWException;
 
+    /**
+     * No-op mapper which can be used as default when no mapper is provided
+     */
+    CloudEventDataMapper<?> NOOP = d -> d;
 }
--- a/api/src/main/java/io/cloudevents/rw/CloudEventReader.java
+++ b/api/src/main/java/io/cloudevents/rw/CloudEventReader.java
@@ -17,7 +17,7 @@
 
 package io.cloudevents.rw;
 
-import io.cloudevents.lang.Nullable;
+import io.cloudevents.CloudEventData;
 
 import javax.annotation.ParametersAreNonnullByDefault;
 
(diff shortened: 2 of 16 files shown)


api/src/main/java/io/cloudevents/rw/CloudEventDataMapper.java
```diff
@@ -37,4 +37,8 @@
      */
     R map(CloudEventData data) throws CloudEventRWException;
 
+    /**
+     * No-op mapper which can be used as default when no mapper is provided
+     */
+    CloudEventDataMapper<?> NOOP = d -> d;
```
eamonnmcmanus, line 43: It might be better for this to be a static method, like [`java.util.function.Function.identity()`](https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/java/util/function/Function.java#L97) and perhaps also called `identity()`:

```java
  static <R extends CloudEventData> CloudEventDataMapper<R> identity() {
    return d -> d;
  }
```

The `<?>` looks like it might make `NOOP` less useful than it could be.

eamonnmcmanus (approved):
Thanks! That's exactly what I had in mind. Just one small suggestion.

api/src/main/java/io/cloudevents/rw/CloudEventDataMapper.java
slinkydeveloper: Sure!

slinkydeveloper: I think it needs to be a wildcard... Otherwise it requires a cast of `d` to `R`

slinkydeveloper: Check it out now

api/src/main/java/io/cloudevents/rw/CloudEventDataMapper.java
```diff
@@ -37,4 +37,10 @@
      */
     R map(CloudEventData data) throws CloudEventRWException;
 
+    /**
+     * No-op identity mapper which can be used as default when no mapper is provided.
+     */
+    static CloudEventDataMapper<? extends CloudEventData> identity() {
```
eamonnmcmanus, line 43: Yeah, I hadn't noticed the problem with the type bound. I guess this is fine. The bound `extends CloudEventData` is redundant because it's already present in the declaration of `CloudEventDataMapper`. In other words just `<?>` would have the same effect.

If the wildcard does cause problems it's possible to restore the `<R>` solution using casts:

```java
    @SuppressWarnings("unchecked")
    static <R extends CloudEventData> CloudEventDataMapper<R> identity() {
      return d -> (R) d;
    }
```

eamonnmcmanus: By the way this suggests to me that the types here aren't really sound. This is saying that you can take any CloudEventData and convert it to MyCloudEventData with an identity transformation. I wonder if the `<R>` parameter is really useful.

slinkydeveloper: > I wonder if the <R> parameter is really useful.

I can show you a use case when this is useful: https://github.com/cloudevents/sdk-java/pull/288/files#diff-d32fb9f764f16209b2cb69f1239b4309975c5601affe825cc4ff9428eca42d1bR52

The `CloudEventDataMapper` can be used directly by the user, when interacting with the event. In this use case, invoking the mapper helper, you get back the concrete type extending `CloudEventData`. Or you can use the mapper inside a `CloudEventReader` or `EventFormat`, and you won't care about the returned concrete type at all because in any case, you assign back to a `CloudEventData`.

slinkydeveloper: Maybe in the identity case `R` should just be `CloudEventData`? like:

```java
    static CloudEventDataMapper<CloudEventData> identity() {
      return d -> d;
    }
```

eamonnmcmanus: Yes! That's obviously the right thing to do. Sorry I made it more complicated.

slinkydeveloper: ok can you review it again? If it's fine, i'm gonna merge it :smile:

eamonnmcmanus (approved):

PR #289, opened 2020-11-19
Move PojoCloudEventData in core

slinkydeveloper:
This PR moves `PojoCloudEventData` from Jackson in core and generalize it, allowing users to carry their pojos as CloudEvent data. 

You can use `PojoCloudEventData#wrap` to wrap the pojo, together with the marshalling function.

For example you can do `PojoCloudEventData.wrap(myPojo, objectMapper::writeValueAsBytes)` to wrap `myPojo` in a `CloudEventData`, expliciting the `objectMapper::writeValueAsBytes` as the method that will serialize the pojo, if needed, back to bytes. 

For a sample use case, look at: https://github.com/cloudevents/sdk-java/pull/288/files#diff-d32fb9f764f16209b2cb69f1239b4309975c5601affe825cc4ff9428eca42d1bR64 


slinkydeveloper:
Related PR https://github.com/cloudevents/sdk-java/pull/291

core/src/main/java/io/cloudevents/core/data/PojoCloudEventData.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Interface defining a conversion from T to byte array. This is similar to {@link java.util.function.Function}
+     * but it allows checked exceptions.
+     *
+     * @param <T> the source type of the conversion
+     */
+    @FunctionalInterface
+    public interface ToBytes<T> {
+        byte[] convert(T data) throws Exception;
```
johanhaleby, line 18: Isn't this unnecessary? Why not just take a function, `Function<T, byte[]>`?

core/src/main/java/io/cloudevents/core/data/PojoCloudEventData.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getValue());
+    }
+
+    /**
+     * Wrap the provided data in a {@link PojoCloudEventData} serializable by the provided mapper.
+     */
+    public static <T> PojoCloudEventData<T> wrap(T data, ToBytes<T> mapper) {
```
johanhaleby, line 71: Personally, I prefer, `from` instead of `wrap`, but it's a minor detail. Also, why not just expose the constructor?

johanhaleby:
I've added some comments, otherwise I think this would be a nice contribution.

core/src/main/java/io/cloudevents/core/data/PojoCloudEventData.java
slinkydeveloper: This is necessary because of checked exceptions, `objectMapper.writeValueAsBytes` gets a checked

core/src/main/java/io/cloudevents/core/data/PojoCloudEventData.java
slinkydeveloper: `wrap` sounds more appropriate because to me `from` implies some conversion, while here we're not performing any conversion, just wrapping one object into another.

The reason to use static methods is only style, I can expose the constructor too, although i prefer to keep it as an impl detail

core/src/main/java/io/cloudevents/core/data/PojoCloudEventData.java
slinkydeveloper: Maybe another name is better?

johanhaleby: Ahh, I didn't notice the `throws`. Classic! 

Personally, I'm one of those that would've preferred if Jackson doesn't throw checked exceptions. But I assume that the checked exception is not arbitrary. You could quite easily wrap a `Function<T, byte[]` in an unchecked function by using a library (vavr has support for this for example) or writing it yourself. For example, the `JsonCloudEventData` (if it still exists?) could provide the `ToBytes` interface because it's a nice utility imo when working with Jackson. But I'm not totally convinced that it generalizes to all cases and that it should be a part of core?

slinkydeveloper: So are you proposing to provide a wrapper in the jackson module to wrap `writeValueAsBytes` and then pass it to this `wrap` method? What if other mappers from other libs have the same issue with checked exceptions too? And, is this overload to the user really worth it, compared to having this interface here?

core/src/main/java/io/cloudevents/core/data/PojoCloudEventData.java
johanhaleby: Wrap is probably fine :) Doesn't matter much to me.

core/src/main/java/io/cloudevents/core/data/PojoCloudEventData.java
johanhaleby: I'm not sure what I'm proposing, I'm thinking out loud as usual :) 

But yeah, I guess one proposal would be to have the `wrap` method in `JsonCloudEvent` accepting `ToBytes` instead of a `Function<T, byte[]>`. 

Whether it's worth it or not is hard to say. The drawback of having a `ToBytes` interface for libraries that doesn't need it is that you lose the nice default methods in "Function", e.g. the ability to compose with other functions etc.

slinkydeveloper: Do we have a use case where, in this particular `wrap` invocation, it might be nice for the user to leverage `Function` composition utilities?

slinkydeveloper: > But yeah, I guess one proposal would be to have the wrap method in JsonCloudEvent accepting ToBytes instead of a Function<T, byte[]>.

So I guess this would look like:

```java
PojoCloudEventData.wrap(pojo, JacksonUtils.bytesConverted(mapper))
```

Or maybe :

```java
PojoCloudEventData.wrap(pojo, SomeUtilsClass.checkedMapper(mapper::writeValueAsBytes))
```

johanhaleby: I can't think of any use case off the top of my head, but personally, I use these things quite frequently. But I doubt that the average Java developer does? 

Yes, I think this is what it would look like:

```
PojoCloudEventData.wrap(pojo, SomeUtilsClass.checkedMapper(mapper::writeValueAsBytes))
```

Or in the case of Jackson (given that `JsonCloudEventData` is still a thing) you could do:

```
JsonCloudEventData.wrap(pojo, mapper::writeValueAsBytes)
```

since here it would arguably be more legitimate to use a `ToBytes` interface? 

But I'm honestly very uncertain if it's worth the extra complexity. It's a personal preference, but if I'm hesitant, I usually pick the [simplest](https://www.infoq.com/presentations/Simple-Made-Easy/) option (i.e. allow building larger things from smaller parts). But I can very well imagine that going with a `ToBytes` interface will be the most convenient option for most users, so you decide :)

slinkydeveloper: > (given that JsonCloudEventData is still a thing)

`JsonCloudEventData` is a separate thing, it's meant to carry `JsonNode` specifically

> But I can very well imagine that going with a ToBytes interface will be the most convenient option for most users, so you decide :)

I think i want to go with this `ToBytes` option, it just seems better to me in this context :smile: We could revisit later

johanhaleby: > JsonCloudEventData is a separate thing, it's meant to carry JsonNode specifically

Sorry, forgot. 

> I think i want to go with this ToBytes option, it just seems better to me in this context 😄 We could revisit later

Sounds good! :)

PR #291, opened 2020-11-19
Add method `BytesCloudEventData#wrap`

slinkydeveloper:
This is done to be consistent with #289. Also deprecating the constructor to force users to use the static method


pierDipi (approved):
It makes sense to me.

LGTM.

PR #293, opened 2020-11-20
Initial introduction of spring support and samples

olegz:
This PR is the result of collaboration between Dave Syer(@dsyer ) and Oleg Zhurakousky(@olegz ) of Spring team (VMWARE).

It provides initial set of utility classes and interfaces to simplify integration of Cloud Events across many Spring projects especially those that are based on [Spring Messaging](https://github.com/spring-projects/spring-framework/tree/master/spring-messaging).
It also provides an initial set of examples to demonstrate several web solutions (i.e., spring-cloud-function-web, spring-webmvc, spring-webflux) as well as streaming solution such as Kafka (but not limited to given an existing Spring binding support for many [messaging brokers](https://spring.io/projects/spring-cloud-stream). More samples will come once the PR is accepted.

IMPORTANT: 
At the moment there is an [experimental support for Cloud Event in Spring Cloud Function](https://github.com/spring-cloud/spring-cloud-function/tree/master/spring-cloud-function-samples/function-sample-cloudevent) which was recently released with version 3.1.0-M5. 
However, there is also a [cloudevents](https://github.com/spring-cloud/spring-cloud-function/tree/cloudevents) branch in Spring Cloud Function, which removes that support and instead relies on Java SDK - this PR.
So in a way this PR (to work properly) is contingent upon ‘cloudevents’ branch of Spring Cloud Function. That said, once this PR is merged the ‘cloudevents’ branch will also be merged to master. 

Signed-off-by: Oleg Zhurakousky <ozhurakousky@vmware.com>

### DIFF
--- a/spring/README.md
+++ b/spring/README.md
@@ -1,0 +1,12 @@
+## Spring Support
+
+### Introduction
+
+This module provides classes and interfaces that can be used by [Spring frameworks](https://spring.io/) and integrations to assist with Cloud Event processing. 
+
+Given that Spring defines [Message](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/messaging/Message.html) abstraction, 
+which perfectly maps to the structure defined by Cloud Events specification, one may say Cloud Events are already supported by any Spring framework that 
+relies on `Message`. So this modules provides several utilities and strategies to simplify working with Cloud Events in the context of Spring 
+frameworks and integrations (see individual component's javadocs for more details).
+
+Please see individual samples in `examples/spring` directory of this SDK for more details.
--- a/spring/pom.xml
+++ b/spring/pom.xml
@@ -1,0 +1,131 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2018-Present The CloudEvents Authors
+  ~ <p>
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~ <p>
+  ~ http://www.apache.org/licenses/LICENSE-2.0
+  ~ <p>
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  ~
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.cloudevents</groupId>
+        <artifactId>cloudevents-parent</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>cloudevents-spring</artifactId>
+    <name>CloudEvents - support for Spring</name>
+    <packaging>jar</packaging>
+
+    <properties>
+        <module-name>io.cloudevents.spring</module-name>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-dependencies</artifactId>
+                <version>2.4.0</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-webmvc</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-messaging</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-function-core</artifactId>
+            <version>3.1.0-SNAPSHOT</version>
+           <optional>true</optional> 
+            <scope>compile</scope>
+        </dependency>
+
+        <!-- Test deps -->
+        <dependency>
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-core</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-core</artifactId>
+            <classifier>tests</classifier>
+            <type>test-jar</type>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+    
+    <repositories>
+        <repository>
+            <id>spring-snapshots</id>
+            <name>Spring Snapshots</name>
+            <url>https://repo.spring.io/libs-snapshot-local</url>
+            <snapshots>
+                <enabled>true</enabled>
+            </snapshots>
+            <releases>
+                <enabled>false</enabled>
+            </releases>
+        </repository>
+        <repository>
+            <id>spring-milestones</id>
+            <name>Spring Milestones</name>
+            <url>https://repo.spring.io/libs-milestone-local</url>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+        </repository>
+        <repository>
+            <id>spring-releases</id>
+            <name>Spring Releases</name>
+            <url>https://repo.spring.io/release</url>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+        </repository>
+    </repositories>
+    
+</project>
+
--- a/spring/src/test/java/io/cloudevents/spring/messaging/CloudEventMessageUtilsTests.java
+++ b/spring/src/test/java/io/cloudevents/spring/messaging/CloudEventMessageUtilsTests.java
@@ -1,0 +1,204 @@
+/*
+ * Copyright 2020-Present The CloudEvents Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cloudevents.spring.messaging;
+
+import java.net.URI;
+import java.util.Map;
+
+import io.cloudevents.core.v1.CloudEventBuilder;
+import io.cloudevents.spring.core.CloudEventAttributeUtils;
+import io.cloudevents.spring.core.MutableCloudEventAttributes;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.converter.MappingJackson2MessageConverter;
+import org.springframework.messaging.support.MessageBuilder;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ *
+ * @author Oleg Zhurakousky
+ *
+ */
+public class CloudEventMessageUtilsTests {
+
+	String payloadWithHttpPrefix = "{\n" +
+	        "    \"ce-specversion\" : \"1.0\",\n" +
+	        "    \"ce-type\" : \"org.springframework\",\n" +
+	        "    \"ce-source\" : \"https://spring.io/\",\n" +
+	        "    \"ce-id\" : \"A234-1234-1234\",\n" +
+	        "    \"ce-datacontenttype\" : \"application/json\",\n" +
+	        "    \"ce-data\" : {\n" +
+	        "        \"version\" : \"1.0\",\n" +
+	        "        \"releaseName\" : \"Spring Framework\",\n" +
+	        "        \"releaseDate\" : \"24-03-2004\"\n" +
+	        "    }\n" +
+	        "}";
+
+	String payloadNoPrefix = "{\n" +
+	        "    \"specversion\" : \"1.0\",\n" +
+	        "    \"type\" : \"org.springframework\",\n" +
+	        "    \"source\" : \"https://spring.io/\",\n" +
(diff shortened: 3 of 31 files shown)


olegz:
@salaboy this is what @dsyer had mentioned earlier

spring/src/test/java/io/cloudevents/spring/messaging/CloudEventMessageUtilsTests.java
```diff
@@ -0,0 +1,204 @@
...
+	        "    \"ce-type\" : \"org.springframework\",\n" +
+	        "    \"ce-source\" : \"https://spring.io/\",\n" +
+	        "    \"ce-id\" : \"A234-1234-1234\",\n" +
+	        "    \"ce-datacontenttype\" : \"application/json\",\n" +
+	        "    \"ce-data\" : {\n" +
+	        "        \"version\" : \"1.0\",\n" +
+	        "        \"releaseName\" : \"Spring Framework\",\n" +
+	        "        \"releaseDate\" : \"24-03-2004\"\n" +
+	        "    }\n" +
+	        "}";
```
slinkydeveloper, line 52: More than hardcoding these, how about putting them in test resources?

spring/README.md
```diff
@@ -0,0 +1,12 @@
+## Spring Support
```
slinkydeveloper, line 1: Can you add the badge and the pom dependency like the other modules? https://github.com/cloudevents/sdk-java/blob/master/kafka/README.md

spring/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+    <properties>
+        <module-name>io.cloudevents.spring</module-name>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-dependencies</artifactId>
+                <version>2.4.0</version>
```
slinkydeveloper, line 42: Can you change this to a property?

spring/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.cloudevents</groupId>
+        <artifactId>cloudevents-parent</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>cloudevents-spring</artifactId>
+    <name>CloudEvents - support for Spring</name>
```
slinkydeveloper, line 30: I think `CloudEvents - Spring` is fine

spring/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-messaging</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-function-core</artifactId>
+            <version>3.1.0-SNAPSHOT</version>
```
slinkydeveloper, line 68: Ah ok so the snapshot is because of what you sad in the comments? Is it possible to merge the support in `spring-cloud-function-core` before merging this one, so we can point out to a proper release? I would love to avoid merging a module depending on snapshots

slinkydeveloper: Also please make this version a property too

spring/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+            </snapshots>
+        </repository>
+        <repository>
+            <id>spring-releases</id>
+            <name>Spring Releases</name>
+            <url>https://repo.spring.io/release</url>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+        </repository>
```
slinkydeveloper, line 127: Aren't these published in maven central too?

slinkydeveloper (commented):
Just did a quick pass, I'll do a proper review on monday

slinkydeveloper (commented):
Don't you convert `Message` <-> `CloudEvent`? Why? I might miss some context here

spring/pom.xml
olegz: yes

spring/src/test/java/io/cloudevents/spring/messaging/CloudEventMessageUtilsTests.java
olegz: Generally I like it that way for tests since it keeps everything in single place. That said I am ok to moving it if you feel strongly about it.

Mrc0113:
+1 on this idea. Once merged I'd like to add a sample using CloudEvents with Spring Cloud Function and the Solace Spring Cloud Stream Binder as a way to exchange Cloud Events over Solace Event Brokers.

olegz:
> Don't you convert `Message` <-> `CloudEvent`? Why? I might miss some context here

@slinkydeveloper  Sorry for the delayed response as this one required a bit more details, so here it is. . .

In SDK `CloudEvent` is a structure to represent the actual Cloud Event. [Spring Message](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/messaging/Message.html) is another structure almost identical to `CloudEvent` structure capable of representing the actual Cloud Event. 
In other words in theory you can use either. 
However, given that many Spring-based frameworks and integrations with Spring are already relying/dependent on Spring `Message` it makes it a natural integration point with Cloud Events since you really don’t need to do much to receive, process and modify Cloud Events (as you can see in current set of examples). Also, as you yourself mentioned several times before, there isn’t a lot of expectation even with non-Spring SDK users to actually use `CloudEvent` type in their actual business applications effectively making `CloudEvent` more of a canonical structure to represent the cloud event within the SDK. That is exactly what Spring `Message` is within Spring.
In short, by relying on Spring `Message` as a structure to represent Cloud Event (and the supporting classes provided as part of this PR) you immediately get integration with any system, framework and application that relies on Spring Messaging.

That said, you question is still valid in the context of interoperability and integration of Spring and non-Spring based systems (which includes Spring frameworks that do not rely on Spring Messaging) at the SDK level and for that we would need to provide something along the lines of [MessageConverter](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/messaging/converter/MessageConverter.html) (standard Spring mechanism for Message type conversion) to do the to/from `CloudEvent` conversion, which we plan to do.

So while this is an initial PR providing integration with several framework from the Spring portfolio, it is hopefully not the last one.

slinkydeveloper:
Thanks for the explanation, I have a couple of comments:

> However, given that many Spring-based frameworks and integrations with Spring are already relying/dependent on Spring Message it makes it a natural integration point with Cloud Events since you really don’t need to do much to receive, process and modify Cloud Events (as you can see in current set of examples). Also, as you yourself mentioned several times before, there isn’t a lot of expectation even with non-Spring SDK users to actually use CloudEvent type in their actual business applications effectively making CloudEvent more of a canonical structure to represent the cloud event within the SDK. That is exactly what Spring Message is within Spring.

Ok I'm trying to look at the examples to grasp how this integration looks like: https://github.com/cloudevents/sdk-java/pull/293/files#diff-3650ef1172cc81f44f7ba479f7399abf655d6ac8b4b88c4f3865652b07c969afR55-R67 <- here what if the user wants to send a custom `CloudEvent`, like defining an extension if something particular is in the input message? Why that function is not `Function<CloudEvent, CloudEvent>`? I would expect the sdk to provide the necessary bits for Spring to go from X to CloudEvent and back, so then Spring can let users write `Function<CloudEvent, CloudEvent>`

> the SDK level and for that we would need to provide something along the lines of MessageConverter (standard Spring mechanism for Message type conversion) to do the to/from CloudEvent conversion, which we plan to do.

I wonder if we should place exactly this inside the sdk, to cover the use case above underlined (if i understood properly how spring works), more than the code provided in this PR, at least as first step. The integrations in this sdk provide the conversion X <-> CloudEvent, in order to allow people to integrate several systems using CloudEvent. That's the major goal of these integrations: then, for each integration, we might add on top the "niceties" to better integrate with the framework (eg in Kafka we provide a `MessageWriter` that provides ProducerRecord <-> CloudEvent, then we provide a `CloudEventSerializer` to play nice with the kafka client).
At least from my understanding of Spring, this PR includes the "niceties" without providing/nor relying on the basic Message <-> CloudEvent conversion, which is the goal of the integration.
Also did you tried to use MessageReader/MessageWriter abstractions to convert back and forth to CloudEvent from Spring's Message? We use these in other integrations (the last merged one is this: https://github.com/cloudevents/sdk-java/tree/master/amqp) and it would be nice to keep everything consistent. I ask because I see a tons of code we could avoid just using them, in particular, seems it's reimplementing the logic behind MessageReader and MessageWriter https://github.com/cloudevents/sdk-java/pull/293/files#diff-f195f970ff73a62bf30fc327ada222ed2a1c5c50f8d0175b0459295162834062. If they're not suitable, can you provide some feedback on how we could improve them?

Last but not least, I might miss something, but in this PR I don't see support nor for structured mode nor for extensions. Our integrations should be 100% compatible and compliant with the CloudEvent spec, so we cannot neglect these 2 features

dsyer:
> I don't see support nor for structured mode nor for extensions

Structured mode is definitely covered (e.g. see the samples - the HTTP one has both binary and structured very explicitly, and the samples in Cloud Function samples support both as well). The presence of the constant `CloudEventAttributeUtils.DATA` is probably a sign that it's there. Extensions are not yet. I think maybe because we don't understand the SDK for them - e.g. the spec says they have the same type system as attributes, but the SDK only supports String, Number and Boolean.

> Also did you tried to use MessageReader/MessageWriter abstractions to convert back and forth to CloudEvent from Spring's Message?

Yes, we looked at that, but it's the wrong level of abstraction really. Spring already handles all those concerns for HTTP, Kafka, etc. So we only really need the CE domain for the attributes (and probably extensions as you point out).

I have an alternative implementation that ships most of the CE domain concerns back to the core SDK: 
https://github.com/dsyer/sdk-java/tree/core. I think I might prefer that approach, but it's kind a a poor fit still because the only object in the CE API that represents both attributes and extensions is `CloudEvent`. It would be a better fit if there was a `CloudEventHeaders` that extended `CloudEventAttributes` and `CloudEventExtensions` (but that might mean quite a big change to core to extract most of `CloudEventBuilder` into a `CloudEventHeadersBuilder`).

> The integrations in this sdk provide the conversion X <-> CloudEvent

That much is clear. But it isn't a very useful model for the majority of Spring applications, mainly (I think) because `CloudEventData` is the wrong abstraction for them. It might be something we could offer (for minority cases), and then you could have your `Function<CloudEvent,CloudEvent>` quite easily. An alternative formulation would be to use `Message<CloudEventData>` (also probably quite easy for a user to implement already). I'm not sure either is very helpful in practice, but there's no reason not to provide the machinery.

Note that if `CloudEventData` was just `Object` we could support it easily and very directly (since it would map directly to `Message` and `HttpEntity` that we already use extensively).

slinkydeveloper:
> Structured mode is definitely covered (e.g. see the samples - the HTTP one has both binary and structured very explicitly, and the Cloud Function samples support both as well)

I'm sorry I didn't saw it! Do you mean this one? https://github.com/cloudevents/sdk-java/pull/293/files#diff-cdd5264380e57fcd33464c1b7992c9d64b37c6b8ca62299fdc154a00cfb9018bR54-R71
Can't the user transparently handle both structured and binary with the same handler? Also Why she/he needs to write 2 handlers for that? As a counter example, here both binary and structured are handled transparently https://github.com/cloudevents/sdk-java/pull/288/files#diff-d32fb9f764f16209b2cb69f1239b4309975c5601affe825cc4ff9428eca42d1bR44
Also, to convert from json to CloudEvent, do you use `EventFormatProvider` (which provides support for potentially other Event formats too like protobuf and avro)?

> Extensions are not covered yet. I think maybe because we don't understand the SDK for them - e.g. the spec says they have the same type system as attributes, but the SDK only supports String, Number and Boolean.

Yes this is kinda a tricky corner case of the spec: extensions supports the same type system of the attributes, but they don't carry the type information with them when they're converted back and forth to protocol binding/event formats (e.g. think about http when the extensions are sent as headers). So, as a general form of abstraction, string/number/boolean is what we support today (which maps more or less to json extension types). But that part of the sdk has definitely room for improvements.

> I have an alternative implementation that ships most of the CE domain concerns back to the core SDK:
https://github.com/dsyer/sdk-java/tree/core. I think I might prefer that approach, but it's kind a a poor fit still because the only object in the CE API that represents both attributes and extensions is CloudEvent. It would be a better fit if there was a CloudEventHeaders that extended CloudEventAttributes and CloudEventExtensions (but that might mean quite a big change to core to extract most of CloudEventBuilder into a CloudEventHeadersBuilder).

So if I understood your point correctly, you're saying that the sdk should not deal with payloads, which are usually handled by other abstractions (like Spring's `Message`), but it should handle only the cloudevent context (attributes and extensions). Am I right? Then how do you deal with unpacking the structured messages transparently then? How do you pass both the context and the payload to another system you're integrating? (say, a library which accepts CloudEvents as input)

> Note that if CloudEventData was just Object (or a type parameter) we could support it easily and very directly (since it would map directly to Message and HttpEntity that we already use extensively).

You can in fact: you could or implement your own CloudEventData which has a method like `get()` that returns object and that's it. The point of `CloudEventData` is to carry any data you want, but just making sure it's serializable back to bytes.

olegz:
> I'm sorry I didn't saw it! Do you mean this one? https://github.com/cloudevents/sdk-java/pull/293/files#diff-cdd5264380e57fcd33464c1b7992c9d64b37c6b8ca62299fdc154a00cfb9018bR54-R71
> Can't the user transparently handle both structured and binary with the same handler? Also Why she/he needs to write 2 handlers for that? As a counter example, here both binary and structured are handled transparently https://github.com/cloudevents/sdk-java/pull/288/files#diff-d32fb9f764f16209b2cb69f1239b4309975c5601affe825cc4ff9428eca42d1bR44
> Also, to convert from json to CloudEvent, do you use `EventFormatProvider` (which provides support for potentially other Event formats too like protobuf and avro)?

That is one of the example using spring-webmvc. No, you don't need to write a different handler between the structured and binary. The examples simply show variety of signatures you can use in thte context of a particular execution and have no bearings on the structure of the incoming event as you can see in [this](https://github.com/olegz/sdk-java/blob/master/examples/spring/cloudevents-spring-function-stream-kafka/src/main/java/io/cloudevents/spring/kafka/CloudeventDemoApplication.java#L62) example.

> Also, to convert from json to CloudEvent, do you use `EventFormatProvider`

In Spring, `MessageConverter` is the standard mechanism to provide type conversion of this sort and we have a large library of existing `MessageConverters` (including Avro, variety of Json and more). Further more it is a natural extension point in practically all frameworks that rely on Spring Messaging, which means you can simply implement `MessageConverter` yourself for some imaginary format, add it as a bean (`@Bean`) and it will be picked up by the framework automatically as you can see in one of the examples that are on the [master of s-c-function](https://github.com/spring-cloud/spring-cloud-function/blob/master/spring-cloud-function-samples/function-sample-cloudevent/src/test/java/io/spring/cloudevent/CloudeventDemoApplicationRESTTests.java#L363). Here you see we have imaginary format with content-type `foo/bar`.

Also, as you said yourself several [month ago](https://github.com/cloudevents/sdk-java/pull/143#issuecomment-621110113):
_...the business logic of an application should not depend on CloudEvent, which is merely a standard exchange format, but on its attributes and payload...._ 
While I am not in the position to argue either side of this argument, I do agree that most of the consumers of Cloud Event need to deal with its payload (data). However some time they may need to have access to meta-data (attributes and extensions) and `Message` structure supports all of these usage patterns. In fact we have identical concerns in Messaging. You may only care about payload - `Function<Foo, Bar>` or if you also care about meta-data `Function<Message<Foo>, Bar>`. The same applies for REST controllers as well as other abstractions. As a user you have flexibility of signatures and 
 underlying frameworks know how the incoming data should be mapped to user types. You can clearly see it in the provided [examples](https://github.com/olegz/sdk-java/blob/master/examples/spring/cloudevents-spring-function-stream-kafka/src/main/java/io/cloudevents/spring/kafka/CloudeventDemoApplication.java#L62). This one is dependent on Spring Cloud Stream receiving events over Kafka (in this example), yet as a user you only care about SpringReleaseEvent (an imaginary POJO) and it is the framework responsibility to serve it to you as you expect while handling all the internals such as type conversion, connectivity to messaging systems, transaction, acks etc,.

> How do you pass both the context and the payload to another system you're integrating?

Spring `Message` defines `payload` (Object) and `headers` (Map<String, Object>). That is what I meant earlier when I said that it is adequately positioned to represent Cloud Event where `data` maps to `payload` and `attributes and extension` map to `headers`.

olegz:
I am closing it it in favor of https://github.com/cloudevents/sdk-java/pull/302 from @dsyer, since it is more alined with the actual SDK implementation.

PR #296, opened 2020-11-23
Introduce CloudEventContext

dsyer:
As discussed in #293 this change introduces a new interface
that extends CloudEventAttributes and CloudEventExtensions.
A builder would be nice, but it's a much bigger change.
Also the BaseBuilder method signatures use CloudEvent when
they could use CloudEventContext, but there's no way to fix
that without major disruption either.

### DIFF
--- a/core/src/main/java/io/cloudevents/core/builder/CloudEventBuilder.java
+++ b/core/src/main/java/io/cloudevents/core/builder/CloudEventBuilder.java
@@ -17,16 +17,18 @@
 
 package io.cloudevents.core.builder;
 
+import java.net.URI;
+import java.time.OffsetDateTime;
+
+import javax.annotation.Nonnull;
+import javax.annotation.ParametersAreNullableByDefault;
+
 import io.cloudevents.CloudEvent;
+import io.cloudevents.CloudEventContext;
 import io.cloudevents.CloudEventData;
 import io.cloudevents.Extension;
 import io.cloudevents.SpecVersion;
 import io.cloudevents.rw.CloudEventWriter;
-
-import javax.annotation.Nonnull;
-import javax.annotation.ParametersAreNullableByDefault;
-import java.net.URI;
-import java.time.OffsetDateTime;
 
 /**
  * Builder interface to build a {@link CloudEvent}.
@@ -273,6 +275,24 @@ public interface CloudEventBuilder extends CloudEventWriter<CloudEvent> {
                 return CloudEventBuilder.v1(event);
             case V03:
                 return CloudEventBuilder.v03(event);
+        }
+        throw new IllegalStateException(
+            "The provided spec version doesn't exist. Please make sure your io.cloudevents deps versions are aligned."
+        );
+    }
+
+    /**
+     * Create a new builder starting from the values of the provided context.
+     *
+     * @param context context to copy values from
+     * @return the new builder
+     */
+    static CloudEventBuilder fromContext(@Nonnull CloudEventContext context) {
+        switch (context.getSpecVersion()) {
+            case V1:
+            return new io.cloudevents.core.v1.CloudEventBuilder(context);
+            case V03:
+            return new io.cloudevents.core.v03.CloudEventBuilder(context);
         }
         throw new IllegalStateException(
             "The provided spec version doesn't exist. Please make sure your io.cloudevents deps versions are aligned."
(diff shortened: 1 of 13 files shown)


slinkydeveloper:
> A builder would be nice, but it's a much bigger change.

Agree and TBH I think this is fine, sdk users expect to build a `CloudEvent`, so I think this is fine

core/src/main/java/io/cloudevents/core/v03/CloudEventBuilder.java
```diff
@@ -50,6 +50,20 @@ public CloudEventBuilder(io.cloudevents.CloudEvent event) {
...
+            if (!name.equals(CloudEventV03.SPECVERSION)) {
+                Object value = context.getAttribute(name);
+                if (value instanceof String) {
+                    withAttribute(name, (String) value);
+                } else {
+                    withAttribute(name, value.toString());
+                }
+            }
+        }
+    }
```
slinkydeveloper, line 65: Could this live in the base builder?

dsyer: I guess it could. But you couldn't use the static constant in that case. If that's OK with you I can make that change (or you can).

core/src/main/java/io/cloudevents/core/impl/BaseCloudEventBuilder.java
```diff
@@ -43,6 +46,32 @@ public BaseCloudEventBuilder() {
...
+            Object value = context.getExtension(name);
+            if (value instanceof String) {
+                withExtension(name, (String) value);
+            } else if (value instanceof Number) {
+                withExtension(name, (Number) value);
+            } else if (value instanceof Boolean) {
+                withExtension(name, (Boolean) value);
+            }
+        }
+    }
```
slinkydeveloper, line 73: Can't you tweak a bit `setAttributes` for that? You don't really need new code here i guess

slinkydeveloper: `setAttributes` is taking the `CloudEvent` just because we didn't had a unique interface for attributes and extensions, but it's doing pretty much what you're doing here

dsyer: `setAttributes()` takes a `CloudEvent` and delegates to some other code. I don't see an obvious way to adapt that to a `CloudEventContext`. If you can suggest something, go ahead.

dsyer: I.e. `setAttributes()` delegates to a `CloudEventReaderAdapter` and that can only be created with a `CloudEvent`.

dsyer: I can add another adapter I guess. Trying that now...

dsyer: See ecfab75

core/src/main/java/io/cloudevents/core/impl/CloudEventReaderAdapter.java
```diff
@@ -25,8 +25,11 @@
```
slinkydeveloper, line 25: You can improve this even further and remove boilerplate subclassing `CloudEventContextReaderAdapter`:

```java
public class CloudEventReaderAdapter extends CloudEventContextReaderAdapter implements CloudEventReader
```

slinkydeveloper (commented):
Just a minor nit, but looks good!

core/src/main/java/io/cloudevents/core/impl/CloudEventReaderAdapter.java
dsyer: I prefer composition over inheritance but it's your choice (236fff0).

core/src/main/java/io/cloudevents/core/builder/CloudEventBuilder.java
```diff
@@ -279,4 +281,22 @@ static CloudEventBuilder from(@Nonnull CloudEvent event) {
...
+            case V1:
+            return new io.cloudevents.core.v1.CloudEventBuilder(context);
+            case V03:
+            return new io.cloudevents.core.v03.CloudEventBuilder(context);
+        }
+        throw new IllegalStateException(
+            "The provided spec version doesn't exist. Please make sure your io.cloudevents deps versions are aligned."
+        );
+    }
+
```
pierDipi, line 301: This method is unused, can we remove it?

slinkydeveloper: It's not unused, it's a new api

pierDipi: Ok, cool!

PR #302, opened 2020-11-24
Alternative implementation of initial introduction of spring support

dsyer:
This PR is the result of collaboration between Dave Syer(@dsyer) and Oleg Zhurakousky(@olegz) of Spring team (VMware).

It provides initial set of utility classes and interfaces to simplify integration of Cloud Events across many Spring projects

This PR duplicates and extends #293. It's an alternative implementation using the core java Cloud Events SDK. 
The examples have been omitted for brevity (and to prevent issues arising with snapshots and Spring Cloud Function).
Of interest is probably the new features that allow `@MVC` and WebFlux users to interact directly with `CloudEvent` if the choose to. Example (as per the tests in the spring module):

```java
		@PostMapping("/event")
		public CloudEvent ce(@RequestBody CloudEvent event) {
			CloudEvent attributes = CloudEventBuilder.from(event).withId(UUID.randomUUID().toString())
					.withSource(URI.create("https://spring.io/foos")).withType("io.spring.event.Foo")
					.withData(event.getData().toBytes()).build();
			return attributes;
		}
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,6 +74,7 @@
         <module>http/vertx</module>
         <module>http/restful-ws</module>
         <module>kafka</module>
+        <module>spring</module>
     </modules>
 
     <properties>
--- a/spring/README.md
+++ b/spring/README.md
@@ -1,0 +1,12 @@
+## Spring Support
+
+### Introduction
+
+This module provides classes and interfaces that can be used by [Spring frameworks](https://spring.io/) and integrations to assist with Cloud Event processing. 
+
+Given that Spring defines [Message](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/messaging/Message.html) abstraction, 
+which perfectly maps to the structure defined by Cloud Events specification, one may say Cloud Events are already supported by any Spring framework that 
+relies on `Message`. So this modules provides several utilities and strategies to simplify working with Cloud Events in the context of Spring 
+frameworks and integrations (see individual component's javadocs for more details).
+
+Please see individual samples in `examples/spring` directory of this SDK for more details.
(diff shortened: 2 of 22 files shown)


slinkydeveloper:
@dsyer after your pr https://github.com/cloudevents/sdk-java/pull/296 and the discussions we had in #293, I think I have an idea to solve the problem of reading/writing only the context attributes... I'll get back to this PR ASAP with updates about that

slinkydeveloper:
Look at this one https://github.com/cloudevents/sdk-java/pull/309

dsyer:
I guess I'll wait for #309 to be merged then.

slinkydeveloper:
@dsyer we don't have to, I'm wondering if that might help the code introduced by this PR too

slinkydeveloper:
My concern with this PR is mostly the introduction of `MutableCloudEventHeaders`, which is spec unversioned and might have bugs related to the various spec version differences... It sounds to me it's kinda duplicating the work already done inside cloudevents-core to properly support attributes/extensions and making sure everything is aligned with the various spec version.
If you want to expose to your users just `CloudEventContext` because the abstraction `CloudEventData` doesn't fit your needs, why don't you build a `CloudEvent` without adding any data and then you just return `CloudEventContext`? Do you really need to go down the path of reimplementing the context? Thanks to #309 this should be even simpler.

Also do we really need a mutable context?

I see the point of "our users needs only the context" and I definitely agree with this vision, so I wonder if we can have a single context implementation (with related builder) that just works for everybody more than duplicating that code... How does that sound to you?

slinkydeveloper:
Do we still need this one?

dsyer:
No, this is superseded now.

PR #306, opened 2020-11-26
Fix NPE on null body with structured message

slinkydeveloper:
Last minute fix before the RC1

### DIFF
--- a/http/vertx/src/test/java/io/cloudevents/http/vertx/VertxMessageFactoryTest.java
+++ b/http/vertx/src/test/java/io/cloudevents/http/vertx/VertxMessageFactoryTest.java
@@ -25,6 +25,7 @@ import io.cloudevents.core.mock.CSVFormat;
 import io.cloudevents.types.Time;
 import io.vertx.core.MultiMap;
 import io.vertx.core.buffer.Buffer;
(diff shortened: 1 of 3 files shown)


http/vertx/src/test/java/io/cloudevents/http/vertx/VertxMessageFactoryTest.java
```diff
@@ -33,9 +34,20 @@
...
 public class VertxMessageFactoryTest {
 
+    @Test
+    public void structuredMessageWithoutBody() {
+        MultiMap headers = MultiMap
+            .caseInsensitiveMultiMap()
+            .add("content-type", CSVFormat.INSTANCE.serializedContentType() + "; charset=utf8");
+
+        assertThatCode(() -> VertxMessageFactory.createReader(headers, null))
+            .hasMessageContaining("null body");
```
pierDipi, line 48: Maybe check the type rather than the message?

slinkydeveloper: The type is CloudEventRWException, which wraps an illegal state exception... I can do it, not sure how much value it adds

pierDipi (approved):

PR #309, opened 2020-11-26
Simplify the Reader/Writer implementations, reducing the knowledge of spec details 

slinkydeveloper:
The goal of this PR is to simplify even further the reader/writer APIs, removing the knowledge of `attribute` and `extensions` for them. The reasons are:

* A protocol binding/event format implementation using reader/writer apis doesn't care if a cloudevent field it's an attribute or an extension, this is a matter only of the implementer of `CloudEventContext` and `CloudEventBuilder`.
* Removing this knowledge simplifies the implementations of both readers and writers, except in the core module, when implementing `CloudEvent` and `CloudEventBuilder`. Now 3rd party reader/writer implementers just need to identify if a field is a cloudevent context field (attribute or extension) or not.
* Now `CloudEventBuilder` is the only one who "owns" all the validation logic to build the event properly
* Unifies the type system used by attributes and extensions: now they all support `String`, `Number`, `Boolean`, `URI`, `OffsetDateTime`

The important changes to look at in this PR are in the api module, in the package `io.cloudevents.rw`, and in the core module, in the package `io.cloudevents.core.v1`,`io.cloudevents.core.v03` and `io.cloudevents.core.impl`.

Note: the code in `CloudEvent` and `CloudEventBuilder` it's not really new, it's just awfully cloned... In followups I plan to improve it, trying to dedup it and organize it better.


matejvasek:
Nice to see this simplification.

johanhaleby:
I agree that it's a good way to be able to access extensions/attributes in a uniform way and that types (String, Number, Boolean, URI, OffsetDateTime) are represented. 

I small thing that came to mind is the name "context attribute", it looks a bit scary to me :). And it could arguably be a bit confusing since it contains the word "attribute", that in this context (if I understand it correctly), should not be confused with "attribute". I would probably prefer "field" or "property" instead.

But all in all, nice work :)

johanhaleby:
@slinkydeveloper Oh, I didn't know that. Then it makes more sense to me. Thanks for the clarification and sorry for the fuss :)

matejvasek (approved):

pierDipi (approved):

PR #312, opened 2020-11-27
HTTP converters for CloudEvent in Spring

dsyer:
Supports MVC and WebFlux (blocking and non-blocking) HTTP.
User can work with `CloudEvent` as a `POJO` type and inject it
into `@ReqestMapping` methods.

Further work is needed to provide support for messaging
and functions.

This is a partial replacement for #302.

### DIFF
--- a/spring/pom.xml
+++ b/spring/pom.xml
@@ -1,0 +1,105 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2018-Present The CloudEvents Authors
+  ~ <p>
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~ <p>
+  ~ http://www.apache.org/licenses/LICENSE-2.0
+  ~ <p>
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  ~
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.cloudevents</groupId>
+        <artifactId>cloudevents-parent</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>cloudevents-spring</artifactId>
+    <name>CloudEvents - support for Spring</name>
+    <packaging>jar</packaging>
+
+    <properties>
+        <module-name>io.cloudevents.spring</module-name>
+        <spring-boot.version>2.4.0</spring-boot.version>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-dependencies</artifactId>
+                <version>${spring-boot.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-webmvc</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-webflux</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-http-basic</artifactId>
+            <version>${project.version}</version>
+            <!-- This will be optional too if we want to support messaging -->
+        </dependency>
+        <dependency>
+            <groupId>org.apache.tomcat.embed</groupId>
+            <artifactId>tomcat-embed-core</artifactId>
+            <optional>true</optional>
+        </dependency>
+
+        <!-- Test deps -->
+        <dependency>
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-json-jackson</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-core</artifactId>
+            <classifier>tests</classifier>
+            <type>test-jar</type>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
+
--- a/spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
+++ b/spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2019-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.cloudevents.spring.http;
+
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+import io.cloudevents.CloudEvent;
+import io.cloudevents.CloudEventContext;
+import io.cloudevents.core.CloudEventUtils;
+import io.cloudevents.core.builder.CloudEventBuilder;
+import io.cloudevents.core.message.MessageReader;
+import io.cloudevents.http.HttpMessageFactory;
+import io.cloudevents.http.impl.HttpMessageWriter;
+
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.ResponseEntity;
+
+/**
+ * Miscellaneous utility methods to assist with Cloud Events in the context of Spring Web
+ * frameworks. Primarily intended for the internal use within Spring-based frameworks or
+ * integrations.
+ *
+ * @author Dave Syer
+ * @since 2.0
+ */
+public class CloudEventHttpUtils {
+
+	private CloudEventHttpUtils() {
+	}
+
+	/**
+	 * Create a {@link MessageReader} to assist in conversion of an HTTP request to a
+	 * {@link CloudEvent}.
+	 * @param headers the HTTP request headers
+	 * @param body the HTTP request body as a byte array
+	 * @return a {@link MessageReader} representing the {@link CloudEvent}
+	 */
+	public static MessageReader toReader(HttpHeaders headers, Supplier<byte[]> body) {
+		return HttpMessageFactory.createReaderFromMultimap(headers, body.get());
+	}
+
+	/**
+	 * Create an {@link HttpMessageWriter} that can hand off a {@link CloudEvent} to an
+	 * HTTP response. Mainly useful in a blocking (not async) setting because the response
+	 * body has to be consumed directly.
+	 * @param headers the response headers (will be mutated)
+	 * @param sendBody a consumer for the response body that puts the bytes on the wire
+	 */
+	public static HttpMessageWriter toWriter(HttpHeaders headers, Consumer<byte[]> sendBody) {
+		return HttpMessageFactory.createWriter(headers::set, sendBody);
+	}
+
+	/**
+	 * Helper method for extracting {@link HttpHeaders} from a {@link CloudEvent}. Can,
+	 * for instance, be used in a <code>&#64;RequestMapping</code> to return a
+	 * {@link ResponseEntity} that has headers copied from a {@link CloudEvent}.
+	 * @param event the input {@link CloudEvent}
+	 * @return the response headers represented by the event
+	 */
+	public static HttpHeaders toHttp(CloudEventContext event) {
+		HttpHeaders headers = new HttpHeaders();
+		CloudEventUtils.toReader(CloudEventBuilder.fromContext(event).build()).read(toWriter(headers, bytes -> {
+		}));
+		return headers;
+	}
+
+	/**
+	 * Helper method for converting {@link HttpHeaders} to a {@link CloudEvent}. The input
+	 * headers must represent a valid event in "binary" form, i.e. it must have headers
+	 * "ce-id", "ce-specversion" etc.
+	 * @param headers the input request headers
+	 * @return a {@link CloudEventBuilder} that can be used to create a new
+	 * {@link CloudEvent}
+	 * 
+	 */
+	public static CloudEventBuilder fromHttp(HttpHeaders headers) {
+		return CloudEventBuilder
+				.fromContext(CloudEventUtils.toEvent(CloudEventHttpUtils.toReader(headers, () -> null)));
+	}
+
+}
--- a/spring/src/main/java/io/cloudevents/spring/http/CloudEventsHeaders.java
+++ b/spring/src/main/java/io/cloudevents/spring/http/CloudEventsHeaders.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2018-Present The CloudEvents Authors
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.cloudevents.spring.http;
+
+import java.util.Map;
+
+import io.cloudevents.core.message.impl.MessageUtils;
+
+import org.springframework.http.HttpHeaders;
+
+public class CloudEventsHeaders {
+
+	public static final String CE_PREFIX = "ce-";
+
+	public static final Map<String, String> ATTRIBUTES_TO_HEADERS = MessageUtils
+			.generateAttributesToHeadersMapping(v -> {
+				if (v.equals("datacontenttype")) {
+					return HttpHeaders.CONTENT_TYPE;
+				}
+				return CE_PREFIX + v;
+			});
+
+	public static final String SPEC_VERSION = ATTRIBUTES_TO_HEADERS.get("specversion");
+
+}
--- a/spring/src/main/java/io/cloudevents/spring/http/package-info.java
+++ b/spring/src/main/java/io/cloudevents/spring/http/package-info.java
@@ -1,0 +1,5 @@
+/**
+ * Provides classes related to working with Cloud Events within the context of Spring and
+ * HTTP.
+ */
+package io.cloudevents.spring.http;
--- a/spring/src/main/java/io/cloudevents/spring/mvc/CloudEventHttpMessageConverter.java
+++ b/spring/src/main/java/io/cloudevents/spring/mvc/CloudEventHttpMessageConverter.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2019-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.cloudevents.spring.mvc;
+
+import java.io.IOException;
+
+import io.cloudevents.CloudEvent;
+import io.cloudevents.core.CloudEventUtils;
+import io.cloudevents.spring.http.CloudEventHttpUtils;
+
+import org.springframework.http.HttpInputMessage;
+import org.springframework.http.HttpOutputMessage;
+import org.springframework.http.MediaType;
+import org.springframework.http.converter.AbstractHttpMessageConverter;
+import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.http.converter.HttpMessageNotReadableException;
+import org.springframework.http.converter.HttpMessageNotWritableException;
+import org.springframework.util.StreamUtils;
+
+/**
+ * An {@link HttpMessageConverter} for {@link CloudEvent CloudEvents}. Supports the use of
+ * {@link CloudEvent} in a <code>&#64;RequestMapping</code> as either a method parameter
+ * or a return value.
+ * 
+ * @author Dave Syer
+ *
+ */
+public class CloudEventHttpMessageConverter extends AbstractHttpMessageConverter<CloudEvent> {
+
+	public CloudEventHttpMessageConverter() {
+		super(MediaType.APPLICATION_OCTET_STREAM, MediaType.ALL);
+	}
+
+	@Override
+	protected boolean supports(Class<?> clazz) {
+		return CloudEvent.class.isAssignableFrom(clazz);
+	}
+
+	@Override
+	protected CloudEvent readInternal(Class<? extends CloudEvent> clazz, HttpInputMessage inputMessage)
+			throws IOException, HttpMessageNotReadableException {
+		byte[] body = StreamUtils.copyToByteArray(inputMessage.getBody());
+		return CloudEventHttpUtils.toReader(inputMessage.getHeaders(), () -> body).toEvent();
+	}
+
+	@Override
+	protected void writeInternal(CloudEvent event, HttpOutputMessage outputMessage)
+			throws IOException, HttpMessageNotWritableException {
+		CloudEventUtils.toReader(event)
+				.read(CloudEventHttpUtils.toWriter(outputMessage.getHeaders(), body -> copy(body, outputMessage)));
+	}
+
+	private void copy(byte[] body, HttpOutputMessage outputMessage) {
+		try {
+			StreamUtils.copy(body, outputMessage.getBody());
+		}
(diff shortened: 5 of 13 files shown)


spring/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+            <artifactId>cloudevents-http-basic</artifactId>
+            <version>${project.version}</version>
+            <!-- This will be optional too if we want to support messaging -->
+        </dependency>
+        <dependency>
+            <groupId>io.cloudevents</groupId>
+            <artifactId>cloudevents-json-jackson</artifactId>
+            <version>${project.version}</version>
+            <optional>true</optional>
+        </dependency>
```
slinkydeveloper, line 76: Do we need this at compile time?

spring/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+    <properties>
+        <module-name>io.cloudevents.spring</module-name>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-dependencies</artifactId>
+                <version>2.4.0</version>
```
slinkydeveloper, line 42: Please make this a property

spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
```diff
@@ -0,0 +1,60 @@
...
+ * @since 2.0
+ */
+public class CloudEventHttpUtils {
+
+	public static MessageReader toReader(HttpHeaders headers, Supplier<byte[]> body) {
+		return HttpMessageFactory.createReaderFromMultimap(headers, body.get());
+	}
+
+	public static HttpMessageWriter toWriter(HttpHeaders headers, Consumer<byte[]> sendBody) {
+		return HttpMessageFactory.createWriter((k, v) -> headers.set(k, v), sendBody);
```
slinkydeveloper, line 45: s/(k, v) -> headers.set(k, v)/headers::set

spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
```diff
@@ -0,0 +1,60 @@
...
+		HttpHeaders headers = new HttpHeaders();
+		CloudEventUtils.toReader(CloudEventBuilder.fromContext(event).build()).read(toWriter(headers, bytes -> {
+		}));
+		return headers;
+	}
+
+	public static CloudEventBuilder fromHttp(HttpHeaders headers) {
+		return CloudEventBuilder
+				.fromContext(CloudEventUtils.toEvent(CloudEventHttpUtils.toReader(headers, () -> null)));
+	}
```
slinkydeveloper, line 58: Since these 2 are used only in tests now, can we keep them inside tests now? If we need them in production code, we can move them later

spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Miscellaneous utility methods to assist with Cloud Events in the context of Spring Web
+ * frameworks. Primarily intended for the internal use within Spring-based frameworks or
+ * integrations.
+ *
+ * @author Dave Syer
+ * @since 2.0
+ */
+public class CloudEventHttpUtils {
```
slinkydeveloper, line 38: This class is usually called `*MessageFactory` in other modules (it's also final and with empty private constructor). Can we rename it to `CloudEventHttpMessageFactory`?

spring/src/main/java/io/cloudevents/spring/http/package-info.java
```diff
@@ -0,0 +1,4 @@
+/**
+ * Provides classes related to working with Cloud Events within the context of Spring MVC.
```
slinkydeveloper, line 2: In this package we handle http generic right? Not only MVC?

spring/src/main/java/io/cloudevents/spring/mvc/CloudEventHttpMessageConverter.java
```diff
@@ -0,0 +1,70 @@
...
+
+	@Override
+	protected boolean supports(Class<?> clazz) {
+		return CloudEvent.class.isAssignableFrom(clazz);
+	}
+
+	@Override
+	protected CloudEvent readInternal(Class<? extends CloudEvent> clazz, HttpInputMessage inputMessage)
+			throws IOException, HttpMessageNotReadableException {
+		byte[] body = StreamUtils.copyToByteArray(inputMessage.getBody());
```
slinkydeveloper, line 50: Is the body always non null?

Also can you move this StreamUtils.copyToByteArray invocation directly in the lambda below?

spring/src/main/java/io/cloudevents/spring/webflux/CloudEventHttpMessageReader.java
```diff
@@ -0,0 +1,68 @@
...
+		HttpHeaders headers = message.getHeaders();
+		Mono<byte[]> body = DataBufferUtils.join(message.getBody()).map(buffer -> {
+			try {
+				return StreamUtils.copyToByteArray(buffer.asInputStream(true));
+			}
+			catch (Exception e) {
+				throw new IllegalArgumentException(e);
+			}
+		});
+		return body.map(bytes -> CloudEventHttpUtils.toReader(headers, () -> bytes)).map(reader -> reader.toEvent());
```
slinkydeveloper, line 65: s/reader -> reader.toEvent()/reader::toEvent

slinkydeveloper, line 65: Does the `map` operation of `Mono` propagate eventual runtime exceptions, right?

spring/src/main/java/io/cloudevents/spring/webflux/CloudEventHttpMessageWriter.java
```diff
@@ -0,0 +1,121 @@
...
+
+	@Override
+	public boolean canWrite(ResolvableType elementType, MediaType mediaType) {
+		return CloudEvent.class.isAssignableFrom(elementType.toClass());
+	}
+
+	@Override
+	public Mono<Void> write(Publisher<? extends CloudEvent> inputStream, ResolvableType elementType,
+			MediaType mediaType, ReactiveHttpOutputMessage message, Map<String, Object> hints) {
+		return Mono.from(inputStream).map(event -> CloudEventUtils.toReader(event))
```
slinkydeveloper, line 56: s/event -> CloudEventUtils.toReader(event)/CloudEventUtils::toReader

spring/src/main/java/io/cloudevents/spring/webflux/CloudEventHttpMessageWriter.java
```diff
@@ -0,0 +1,121 @@
...
+
+		@Override
+		public ReactiveHttpMessageWriter withExtension(String name, String value) throws CloudEventRWException {
+			this.response.getHeaders().set("ce-" + name, value);
+			return this;
+		}
+
+		@Override
+		public Mono<Void> end(CloudEventData value) throws CloudEventRWException {
+			if (this.response.isCommitted()) {
```
slinkydeveloper, line 93: Is there some situation where this might happen? Also why you didn't added to the other `end()` and `setEvent()` too?

spring/src/test/java/io/cloudevents/spring/mvc/MvcRestControllerTests.java
```diff
@@ -0,0 +1,201 @@
...
+				converters.add(0, new CloudEventHttpMessageConverter());
+			}
+
+		}
+
+	}
+
+}
+
+class Foo {
```
slinkydeveloper, line 177: Can you place this in another file?

spring/src/test/java/io/cloudevents/spring/mvc/MvcRestControllerTests.java
```diff
@@ -0,0 +1,201 @@
...
+	@Autowired
+	private TestRestTemplate rest;
+
+	@LocalServerPort
+	private int port;
+
+	@Test
+	void echoWithCorrectHeaders() {
+
+		ResponseEntity<String> response = rest.exchange(RequestEntity.post(URI.create("http://localhost:" + port + "/")) //
```
slinkydeveloper, line 64: Unrelated to this PR: can we use the MessageReader and MessageWriter for the "client" part too? I recall in jaxrs i could reuse them, reimplementing one or two additional interfaces. Is it the case here too?

spring/src/test/java/io/cloudevents/spring/mvc/MvcRestControllerTests.java
```diff
@@ -0,0 +1,201 @@
...
+	@RestController
+	static class TestApplication {
+
+		@PostMapping("/")
+		public ResponseEntity<Foo> echo(@RequestBody Foo foo, @RequestHeader HttpHeaders headers) {
+			CloudEvent attributes = CloudEventHttpUtils.fromHttp(headers).withId(UUID.randomUUID().toString())
+					.withSource(URI.create("https://spring.io/foos")).withType("io.spring.event.Foo").build();
+			HttpHeaders outgoing = CloudEventHttpUtils.toHttp(attributes);
+			return ResponseEntity.ok().headers(outgoing).body(foo);
+		}
```
slinkydeveloper, line 153: Shouldn't you reply back with a `CloudEvent` directly? (and let the framework invoke the writer)

spring/src/test/java/io/cloudevents/spring/mvc/MvcRestControllerTests.java
```diff
@@ -0,0 +1,201 @@
...
+
+		@PostMapping("/event")
+		public CloudEvent ce(@RequestBody CloudEvent event) {
+			CloudEvent attributes = CloudEventBuilder.from(event).withId(UUID.randomUUID().toString())
```
slinkydeveloper, line 157: Can you indent a bit more from the `from` invocation?

spring/src/test/java/io/cloudevents/spring/webflux/WebFluxRestControllerTests.java
```diff
@@ -0,0 +1,201 @@
...
+				configurer.customCodecs().register(new CloudEventHttpMessageWriter());
+			}
+
+		}
+
+	}
+
+}
+
+class Foo {
```
slinkydeveloper, line 177: Since you have this class in the other test package, maybe let's just have one copy of it in some test package, like `io.cloudevents.spring.test` in test classpath?

spring/src/test/java/io/cloudevents/spring/webflux/WebFluxRestControllerTests.java
```diff
@@ -0,0 +1,201 @@
...
+	@SpringBootApplication
+	@RestController
+	static class TestApplication {
+
+		@PostMapping("/")
+		public ResponseEntity<Foo> echo(@RequestBody Foo foo, @RequestHeader HttpHeaders headers) {
+			CloudEvent attributes = CloudEventHttpUtils.fromHttp(headers).withId(UUID.randomUUID().toString())
+					.withSource(URI.create("https://spring.io/foos")).withType("io.spring.event.Foo").build();
+			HttpHeaders outgoing = CloudEventHttpUtils.toHttp(attributes);
+			return ResponseEntity.ok().headers(outgoing).body(foo);
```
slinkydeveloper, line 152: Same concern as before, I may misunderstood the logic of the test, but shouldn't you test here that the response cloudevent is properly written? Why are you manually performing this conversion?

slinkydeveloper (commented):
The overall direction looks very good! I mostly left style comments and some questions about the tests.

Can you push in another PR (no matter if it doesn't build, it's just to figure out how it works) the example that uses the stuff added here? So after we merge it, we can rebase the example PR and merge it as well

spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
dsyer: The tests are more like integration tests illustrating example usage. If we add actual example apps I would add the same code there, so no, I don't think these can be private or test scoped.

spring/src/test/java/io/cloudevents/spring/mvc/MvcRestControllerTests.java
dsyer: Probably. We can add that later I guess.

spring/src/test/java/io/cloudevents/spring/webflux/WebFluxRestControllerTests.java
dsyer: I'm not sure. It's just a demo POJO. There's no reason to share it IMO.

spring/src/test/java/io/cloudevents/spring/mvc/MvcRestControllerTests.java
dsyer: I think so. Not sure what you mean, but see what you think.

spring/src/test/java/io/cloudevents/spring/webflux/WebFluxRestControllerTests.java
dsyer: Because as a user I want to consume a `Foo` (POJO) and `CloudEvent` is not appropriate for that (as we discussed yesterday).

spring/src/main/java/io/cloudevents/spring/webflux/CloudEventHttpMessageReader.java
dsyer: I think you mean `MessageReader::toEvent`.

spring/src/main/java/io/cloudevents/spring/mvc/CloudEventHttpMessageConverter.java
dsyer: The body is never null (according to the javadocs). You can move the `StreamUtils` call into the lambda but then you'd have to try/catch/rethrow `IOException`. I prefer not to.

spring/src/main/java/io/cloudevents/spring/http/package-info.java
dsyer: It's shared between MVC and Webflux, so yes, its generic.

spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
dsyer: I guess. But it isn't just a message reader/writer factory.

spring/src/main/java/io/cloudevents/spring/webflux/CloudEventHttpMessageWriter.java
dsyer: Yes

spring/src/main/java/io/cloudevents/spring/webflux/CloudEventHttpMessageReader.java
dsyer: Yes

spring/src/main/java/io/cloudevents/spring/webflux/CloudEventHttpMessageWriter.java
dsyer: I guess it might be over defensive. I removed this check (it was copied from the vertx example), and I'm pretty sure it will bubble up as a runtime exception if that condition even is met.

spring/src/test/java/io/cloudevents/spring/mvc/MvcRestControllerTests.java
dsyer: Why? It is only used by the test.

spring/src/test/java/io/cloudevents/spring/mvc/MvcRestControllerTests.java
dsyer: That wasn't the intention. I want to reply with a `Foo` and have the framework deal with the serialization.

dsyer:
> Can you push in another PR (no matter if it doesn't build, it's just to figure out how it works) the example that uses the stuff added here?

The tests in this PR represent valid samples, so I think you have everything. I can refactor them into separate examples later if you want.

spring/src/test/java/io/cloudevents/spring/webflux/WebFluxRestControllerTests.java
slinkydeveloper: Ok but then what happens if I send a structured event to `/`? I won't be able to parse it right? That's why I guess here you want to accept a `CloudEvent` and then with `PojoCloudEventDataMapper` extract `Foo` from it

spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
dsyer: I put in the private constructor.

spring/src/test/java/io/cloudevents/spring/webflux/WebFluxRestControllerTests.java
slinkydeveloper: It's just do deduplicate that code

spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
slinkydeveloper: I think this is related to this point here: https://github.com/cloudevents/sdk-java/pull/312#discussion_r531494254

spring/src/test/java/io/cloudevents/spring/webflux/WebFluxRestControllerTests.java
slinkydeveloper: As a user, I don't really care if the CloudEvent is wrapped in a json (structured mode) or smashed in the headers (binary mode), I just want that the framework gives me the entity in a uniform way, then it's up to me to figure out extracting `Foo`. Then, as i mentioned yesterday, in next iterations we might improve it adding an annotation to extract `Foo` straight from `CloudEventData` or something like that.

dsyer: And it is (transparent), if you consume `CloudEvent`, as is show by the tests here. But if you consume `Foo`, currently the best option is to have Spring do the conversion; and then you have to deal explicitly with the headers. What might be useful is to have support for `CloudEventContext` as a pseudonym for `HttpHeaders`, but that's kind of an enhancement on top of what is here.

spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
dsyer: Yes. So as long as we need to support the user doing conversion from `CloudEventContext` to `HttpHeaders` then we need this utility to be public.

spring/src/test/java/io/cloudevents/spring/webflux/WebFluxRestControllerTests.java
dsyer: There's no duplication. The two Foos are not related in any way.

spring/src/test/java/io/cloudevents/spring/webflux/WebFluxRestControllerTests.java
slinkydeveloper: Ok but in that case, the user explicitly accepts to consume only binary CloudEvents. Does this makes sense? Do we have use cases where the user wants to only explicitly accept binary CloudEvents? My question is: do we have a use case where a user would developer an endpoint like this one?

slinkydeveloper: By spec every implementation (senders and receivers) SHOULD support both binary and structured: https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md#13-content-modes

The way I see it is that in any eventual use case the user wants to interact with CloudEvevents, he/she must be able to accept both, otherwise its implementation won't be 100% compliant, causing problems when 2 systems have to communicate.

dsyer: > Do we have a use case where a user would developer an endpoint like this one?

Yes. It's quite possible to implement the complementary endpoint that accepts a `Foo` from a structured request body. I think I even did that in the original #293 PR. We can definitely refine this in the future, but the basic building blocks from the Spring side are already in place (content negotiation and deserialization).

dsyer: > The way I see it is that in any eventual use case the user wants to interact with CloudEvents, he/she must be able to accept both,

That's explicitly what the features in this PR supports directly (per the integration test example).

slinkydeveloper: > It's quite possible to implement the complementary endpoint that accepts a Foo from a structured request body

Why the user should define a complementary endpoint? Isn't that like exposing the difference between binary and structured, forcing the user to expose 2 endpoints for receiving the same entity in 2 different "encodings"?

I think I explained myself badly: I would love to give the user the ability to extract `Foo` from the endpoint signature just like you did in this test, but without exposing the problem binary/structured. Unless I miss something, my understanding here is that the user, to extract that `Foo`, needs to have 2 endpoints one for binary one for structured, am I right?

dsyer: Yes, I believe so. We can maybe try and fix that later, but for now that is how it is (and I for one don't think that's too terrible).

spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
```diff
@@ -0,0 +1,63 @@
...
+
+	private CloudEventHttpUtils() {
+	}
+
+	public static MessageReader toReader(HttpHeaders headers, Supplier<byte[]> body) {
```
slinkydeveloper, line 43: javadoc

spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
```diff
@@ -0,0 +1,63 @@
...
+		return HttpMessageFactory.createReaderFromMultimap(headers, body.get());
+	}
+
+	public static HttpMessageWriter toWriter(HttpHeaders headers, Consumer<byte[]> sendBody) {
```
slinkydeveloper, line 47: javadoc

spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
```diff
@@ -0,0 +1,63 @@
...
+		return HttpMessageFactory.createWriter(headers::set, sendBody);
+	}
+
+	public static HttpHeaders toHttp(CloudEventContext event) {
```
slinkydeveloper, line 51: javadoc

spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
```diff
@@ -0,0 +1,63 @@
...
+		HttpHeaders headers = new HttpHeaders();
+		CloudEventUtils.toReader(CloudEventBuilder.fromContext(event).build()).read(toWriter(headers, bytes -> {
+		}));
+		return headers;
+	}
+
+	public static CloudEventBuilder fromHttp(HttpHeaders headers) {
```
slinkydeveloper, line 58: javadoc

spring/src/main/java/io/cloudevents/spring/mvc/CloudEventHttpMessageConverter.java
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.http.converter.AbstractHttpMessageConverter;
+import org.springframework.http.converter.HttpMessageNotReadableException;
+import org.springframework.http.converter.HttpMessageNotWritableException;
+import org.springframework.util.StreamUtils;
+
+/**
+ * @author Dave Syer
+ *
+ */
+public class CloudEventHttpMessageConverter extends AbstractHttpMessageConverter<CloudEvent> {
```
slinkydeveloper, line 36: Little javadoc that explains what this does?

spring/src/main/java/io/cloudevents/spring/webflux/CloudEventHttpMessageReader.java
```diff
@@ -0,0 +1,69 @@
...
+
+import org.springframework.core.ResolvableType;
+import org.springframework.core.io.buffer.DataBufferUtils;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.MediaType;
+import org.springframework.http.ReactiveHttpInputMessage;
+import org.springframework.http.codec.HttpMessageReader;
+import org.springframework.util.StreamUtils;
+
+public class CloudEventHttpMessageReader implements HttpMessageReader<CloudEvent> {
```
slinkydeveloper, line 36: javadoc?

spring/src/main/java/io/cloudevents/spring/webflux/CloudEventHttpMessageWriter.java
```diff
@@ -0,0 +1,117 @@
...
+import reactor.core.publisher.Mono;
+
+import org.springframework.core.ResolvableType;
+import org.springframework.core.io.buffer.DataBuffer;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.MediaType;
+import org.springframework.http.ReactiveHttpOutputMessage;
+import org.springframework.http.codec.HttpMessageWriter;
+
+public class CloudEventHttpMessageWriter implements HttpMessageWriter<CloudEvent> {
```
slinkydeveloper, line 41: javadoc

slinkydeveloper (commented):
Some javadoc comments and we're good

spring/src/test/java/io/cloudevents/spring/webflux/WebFluxRestControllerTests.java
slinkydeveloper: TBH I think it's a double edged sword, but ok let's expose this stuff now and let's be explicit in the javadoc that this won't work with the structured mode, ok?

slinkydeveloper:
@dsyer after we merge this one, can you provide a pr to add this sample https://github.com/cloudevents/sdk-java/pull/293/files#diff-cdd5264380e57fcd33464c1b7992c9d64b37c6b8ca62299fdc154a00cfb9018b using the converters added here?

dsyer:
> can you provide a pr to add this sample

Yes, of course. I also have support for `Message<?>` on a branch, so there will be another PR for that.

slinkydeveloper (approved):

spring/src/main/java/io/cloudevents/spring/http/CloudEventsHeaders.java
```diff
@@ -0,0 +1,40 @@
...
+
+import io.cloudevents.core.message.impl.MessageUtils;
+
+import org.springframework.http.HttpHeaders;
+
+public class CloudEventsHeaders {
+
+	public static final String CE_PREFIX = "ce-";
+
+	public static final Map<String, String> ATTRIBUTES_TO_HEADERS = MessageUtils
```
rhuss, line 30: minor thing: I could see the maintenance of this mapping as part of the `MessageUtils`, too.

rhuss (commented):
looks good to me on a quick view, with one question on the dependency of spring internal repos.

spring/src/test/java/io/cloudevents/spring/webflux/WebFluxRestControllerTests.java
dsyer: I'm not sure that's accurate. It works fine in structured mode, but you have to be explicit about that in your handler. I'm not sure the Javadoc for the library jar can capture that adequately. What about the README? Or the integration tests?

spring/pom.xml
```diff
@@ -0,0 +1,134 @@
...
+            <artifactId>spring-boot-starter-web</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+    <repositories>
```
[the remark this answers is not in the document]
dsyer: Good catch. They can be removed I think.

spring/src/main/java/io/cloudevents/spring/http/CloudEventsHeaders.java
dsyer: Is there some action required in the PR? Not sure what you mean really.

spring/src/test/java/io/cloudevents/spring/webflux/WebFluxRestControllerTests.java
dsyer: BTW the javadocs for `HttpUtils` already has this:

> The input headers must represent a valid event in "binary" form, i.e. it must have headers "ce-id", "ce-specversion" etc.

So that's pretty explicitly referring to binary mode in CE terms.

PR #315, opened 2020-11-30
Add support for Message<CloudEvent>

dsyer:
The integration test demonstrating the usage in Spring Cloud
Function depends on a snapshot, but we can stabilize that quickly
whenever we get a green light on the rest of the PR. (Or we could
just remove it and add it back later as a sample when Spring Cloud
has a release.)


spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
```diff
@@ -0,0 +1,78 @@
...
+
+	@Override
+	public Object fromMessage(Message<?> message, Class<?> targetClass) {
+		byte[] payload = getBinaryData(message);
+		if (CloudEvent.class.isAssignableFrom(targetClass)
+				&& message.getHeaders().containsKey(CloudEventsHeaders.SPEC_VERSION)) {
+			CloudEventBuilder builder = CloudEventHeaderUtils.fromMap(message.getHeaders());
+			if (payload != null) {
+				return builder.withData(payload).build();
+			}
```
slinkydeveloper, line 49: Why you build the context with the message reader but the payload not?

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	@Override
+	public Message<?> toMessage(Object payload, MessageHeaders headers) {
+		if (payload instanceof CloudEvent) {
+			CloudEvent event = (CloudEvent) payload;
+			CloudEventData data = event.getData();
+			byte[] bytes = data == null ? new byte[0] : data.toBytes();
+			return MessageBuilder.withPayload(bytes).copyHeaders(headers)
+					.copyHeaders(CloudEventHeaderUtils.toMap(event)).build();
```
slinkydeveloper, line 73: Same question, why do you use the readers/writers just for the context?

slinkydeveloper, line 73: Also, why do you copyHeaders 2 times?

slinkydeveloper (commented):
Can you rebase on master?

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
dsyer: Same answer

dsyer: The incoming message might have headers that are not in the `CloudEventContext`. The first `copyHeaders()` makes sure those are used as well.

dsyer:
> Can you rebase on master?

Done. Looking at your changes now.

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
dsyer: Because the `CloudEventHeaderUtils` is for headers (i.e. `CloudEventContext`) but the core SDK doesn't allow that separation currently. I noticed the new `CloudEventContextWriter`, so I think that can be used instead of `MessageWriter` in the `toMap()` method. The other direction doesn't work.

slinkydeveloper: In this case, what happens if the incoming message is a structured cloudevent message?

dsyer: This doesn't support that use case I guess - it's the same discussion we had about the HTTP converters - Spring has deep knowledge of how to do type conversions and I want to keep that (even if it restricts me to binary cloud event messages). I could look at extending the PR. Or we could put it in a separate one. Which do you prefer?

slinkydeveloper: I would prefer to have it in this PR

dsyer: OK. That was fun (see latest commit).

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
```diff
@@ -0,0 +1,151 @@
...
+		return null;
+	}
+
+	private MessageReader createMessageReader(Message<?> message) {
+		Supplier<String> contentTypeHeaderReader = () -> contentType(message);
+		Function<EventFormat, MessageReader> structuredMessageFactory = format -> structuredMessageReader(message,
+				format);
+		Supplier<String> specVersionHeaderReader = () -> version(message);
+		Function<SpecVersion, MessageReader> binaryMessageFactory = version -> binaryMessageReader(message, version);
+		return MessageUtils.parseStructuredOrBinaryMessage(contentTypeHeaderReader, structuredMessageFactory,
```
slinkydeveloper, line 78: Can you inline all the above lambdas?

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventHeaderUtils.java
```diff
@@ -0,0 +1,127 @@
...
+	 */
+	public static Map<String, ?> toMap(CloudEvent event) {
+		Map<String, Object> headers = new HashMap<>();
+		CloudEventUtils.toContextReader(event).readContext(new MapWriter(headers));
+		// Probably this should be done in CloudEventContextReaderAdapter
+		headers.put(CE_PREFIX + "specversion", event.getSpecVersion().toString());
+		return headers;
+	}
+
+	private static class MapContextMessageReader extends BaseGenericBinaryMessageReaderImpl<String, Object> {
```
slinkydeveloper, line 74: Can you push this in another file? It would help the readability

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventHeaderUtils.java
```diff
@@ -0,0 +1,127 @@
...
+
+		private final Consumer<BiConsumer<String, Object>> forEachHeader;
+
+		public MapContextMessageReader(SpecVersion version, Consumer<BiConsumer<String, Object>> forEachHeader) {
+			super(version, null);
```
slinkydeveloper, line 79: Can we read the body too here? You can always pass null when using it in `toMap()`

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventHeaderUtils.java
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		@Override
+		protected String toCloudEventsValue(Object value) {
+			return value.toString();
+		}
+
+	}
+
+	private static class MapWriter implements CloudEventContextWriter {
```
slinkydeveloper, line 111: Can you push this class outside this file too?

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
slinkydeveloper: I think you should rather adapt the `MapWriter` to be a `BinaryMessageWriter` and then use it here, and in `toMap` you use the `MapWriter` without empty body

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	@Override
+	public Message<?> toMessage(Object payload, MessageHeaders headers) {
+		if (payload instanceof CloudEvent) {
+			CloudEvent event = (CloudEvent) payload;
+			CloudEventData data = event.getData();
+			byte[] bytes = data == null ? new byte[0] : data.toBytes();
+			return MessageBuilder.withPayload(bytes).copyHeaders(headers)
+					.copyHeaders(CloudEventHeaderUtils.toMap(event)).build();
```
slinkydeveloper, line 63: Here you can just use the `MapContextMessageWriter` at this point, right?

```
CloudEventUtils.toReader(event).read(new MapContextMessageWriter(headers::put, MessageBuilder::withPayload))
```

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
```diff
@@ -0,0 +1,120 @@
...
+		}
+		return null;
+	}
+
+	private MessageReader binaryMessageReader(Message<?> message, SpecVersion version) {
+		return new MapContextMessageReader(version, message.getHeaders()::forEach,
+				BytesCloudEventData.wrap(getBinaryData(message)));
+	}
+
+	private MessageReader structuredMessageReader(Message<?> message, EventFormat format) {
```
slinkydeveloper, line 89: No need to create a new class for that, you can use https://github.com/cloudevents/sdk-java/blob/master/core/src/main/java/io/cloudevents/core/message/impl/GenericStructuredMessageReader.java#L49

spring/src/main/java/io/cloudevents/spring/messaging/MapContextMessageReader.java
```diff
@@ -0,0 +1,74 @@
...
+import static io.cloudevents.spring.messaging.CloudEventsHeaders.CE_PREFIX;
+import static org.springframework.messaging.MessageHeaders.CONTENT_TYPE;
+
+/**
+ * Utility for converting maps (message headers) to `CloudEvent` contexts.
+ * 
+ * @author Dave Syer
+ *
+ */
+class MapContextMessageReader extends BaseGenericBinaryMessageReaderImpl<String, Object> {
```
slinkydeveloper, line 34: Cab we have another name here? Like `SpringMessageReader` or something like that? `MapContext` is confusing to me because it's not only about a map and about context.

spring/src/main/java/io/cloudevents/spring/messaging/MapContextMessageWriter.java
```diff
@@ -0,0 +1,82 @@
...
+import io.cloudevents.rw.CloudEventWriter;
+
+/**
+ * Internal utility class for copying <code>CloudEvent</code> context to a map (message
+ * headers).
+ * 
+ * @author Dave Syer
+ *
+ */
+class MapContextMessageWriter implements CloudEventWriter<Void>, MessageWriter<MapContextMessageWriter, Void> {
```
slinkydeveloper, line 36: Same concern as above

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventHeaderUtils.java
```diff
@@ -0,0 +1,68 @@
...
+	/**
+	 * Helper method for converting {@link MessageHeaders} to a {@link CloudEvent}. The
+	 * input headers must represent a valid event in "binary" form, i.e. it must have
+	 * headers "ce-id", "ce-specversion" etc.
+	 * @param headers the input request headers
+	 * @return a {@link CloudEventBuilder} that can be used to create a new
+	 * {@link CloudEvent}
+	 * 
+	 */
+	public static CloudEventBuilder fromMap(Map<String, ?> headers) {
```
slinkydeveloper, line 48: I don't really expect a builder as return value, i expect a `CloudEventContext`

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventsHeaders.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.cloudevents.spring.messaging;
+
+public class CloudEventsHeaders {
+
+	public static final String CE_PREFIX = "ce-";
```
slinkydeveloper, line 22: This prefix doesn't work for all transports, how do we deal with that?

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventHeaderUtils.java
```diff
@@ -0,0 +1,68 @@
...
+		Object value = headers.get(CloudEventsHeaders.SPEC_VERSION);
+		SpecVersion version = value == null ? SpecVersion.V1 : SpecVersion.parse(value.toString());
+		return CloudEventBuilder
+				.fromContext(CloudEventUtils.toEvent(new MapContextMessageReader(version, headers::forEach)));
+	}
+
+	/**
+	 * Helper method for extracting {@link MessageHeaders} from a {@link CloudEvent}.
```
slinkydeveloper, line 56: Explicit here that the headers will be serialized as binary mode

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventHeaderUtils.java
```diff
@@ -0,0 +1,68 @@
...
+	 * @param event the input {@link CloudEvent}
+	 * @return the response headers represented by the event
+	 */
+	public static Map<String, ?> toMap(CloudEvent event) {
+		Map<String, Object> headers = new HashMap<>();
+		CloudEventUtils.toContextReader(event).readContext(new MapContextMessageWriter(headers::put));
+		// Probably this should be done in CloudEventContextReaderAdapter
+		headers.put(CE_PREFIX + "specversion", event.getSpecVersion().toString());
```
slinkydeveloper, line 64: Can you move this before invoking readContext?

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventHeaderUtils.java
olegz: This usually would be called by the framework  (e.g., one of Spring frameworks) and it may want to add additional headers, attributes based on configuration, so returning an open builder is a pretty common pattern

spring/src/main/java/io/cloudevents/spring/messaging/MapContextMessageReader.java
olegz: I think bringing 'Spring' into the name would be even more confusing, since there is really nothing specific to Spring here other then `MessageHeaders.CONTENT_TYPE` which could be removed in favor of some local constant. Perhaps `KeyValueMessageReader`?

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventHeaderUtils.java
slinkydeveloper: It's still weird, I've asked for a conversion from A to B, while here it's doing the conversion from A to B and then to a builder, which will return C. It's not symmetric with `toMap`, you can go from B to C easily and also IMO goes a little bit against the [single responsability principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).

spring/src/main/java/io/cloudevents/spring/messaging/MapContextMessageReader.java
slinkydeveloper: Yeah perhaps SpringMessageReader doesn't sound right, `KeyValueMessageReader` doesn't sound right too because here you're handling the body too... Not sure what's the right name tbh, but `MapContextMessageReader` is for sure wrong... Maybe we can live with just `MessageReader`?

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
dsyer: I could. Isn't the result identical? If it is I would prefer to leave it as it is as the result is a `Message` (so starting with `MessageBuilder` makes more sense).

spring/src/main/java/io/cloudevents/spring/messaging/MapContextMessageReader.java
dsyer: I don't care. It was private originally, so the name didn't matter.

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventHeaderUtils.java
dsyer: I think this is a remnant of the code before `CloudEventContext` existed. I can change it, I guess.

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventsHeaders.java
dsyer: It will work for all transports because Spring will convert the transport specific headers to the canonical form. We control that on the Spring side - we can agree that is the contract and it will just work. My next job is to create some samples that demonstrate.

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventHeaderUtils.java
dsyer: It's not precisely "binary mode" because the result is a `Map` not a `CloudEvent`, but I guess we can say that it is "analogous to binary mode" or something.

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventHeaderUtils.java
dsyer: Not sure I understand. You want to call `headers.put()` before `CloudEventUtils.toContextReader(event).readContext()`? Whatever :shrug:.

spring/src/main/java/io/cloudevents/spring/messaging/MapContextMessageReader.java
dsyer: `MessageReader` makes no sense to me though - it duplicates the interface that it implements.

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventHeaderUtils.java
slinkydeveloper: yeah it's logically more correct and clean, although practically there is no difference

spring/src/test/java/io/cloudevents/spring/messaging/CloudEventMessageConverterTests.java
```diff
@@ -0,0 +1,130 @@
...
+		CloudEvent event = (CloudEvent) converter.fromMessage(message, CloudEvent.class);
+		assertThat(event).isNotNull();
+		assertThat(event.getSpecVersion()).isEqualTo(SpecVersion.V1);
+		assertThat(event.getId()).isEqualTo("12345");
+		assertThat(event.getSource()).isEqualTo(URI.create("https://spring.io/events"));
+		assertThat(event.getType()).isEqualTo("io.spring.event");
+	}
+
+	@Test
+	void binaryCloudEventStringPayload() {
```
slinkydeveloper, line 100: this is structured, not binary

spring/src/test/java/io/cloudevents/spring/messaging/CloudEventMessageConverterTests.java
```diff
@@ -0,0 +1,130 @@
...
+		CloudEvent event = (CloudEvent) converter.fromMessage(message, CloudEvent.class);
+		assertThat(event).isNotNull();
+		assertThat(event.getSpecVersion()).isEqualTo(SpecVersion.V1);
+		assertThat(event.getId()).isEqualTo("12345");
+		assertThat(event.getSource()).isEqualTo(URI.create("https://spring.io/events"));
+		assertThat(event.getType()).isEqualTo("io.spring.event");
+	}
+
+	@Test
+	void binaryCloudEvent() {
```
slinkydeveloper, line 87: This is structured, not binary

spring/src/test/java/io/cloudevents/spring/messaging/CloudEventMessageConverterTests.java
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	@Test
+	void targetNotCloudEvent() {
+		Message<?> message = MessageBuilder.withPayload(new byte[0]).build();
+		assertThat(converter.fromMessage(message, String.class)).isNull();
+	}
+
+	@Test
+	void validCloudEvent() {
```
slinkydeveloper, line 74: This is binary

slinkydeveloper (commented):
Small nits

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
```diff
@@ -0,0 +1,124 @@
...
+
+	@Override
+	public Message<?> toMessage(Object payload, MessageHeaders headers) {
+		if (payload instanceof CloudEvent) {
+			CloudEvent event = (CloudEvent) payload;
+			Map<String, Object> map = new HashMap<>();
+			String contentType = contentType(headers);
+			if (contentType == null || EventFormatProvider.getInstance().resolveFormat(contentType) == null) {
+				// binary format
+				map.putAll(CloudEventHeaderUtils.toMap(event));
```
slinkydeveloper, line 62: You don't need this, because it's done already in the reader :smile:

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
```diff
@@ -0,0 +1,124 @@
...
+			}
+			map.putAll(headers);
+			return createMessageReader(event, headers).read(new MessageBuilderMessageWriter(map));
```
slinkydeveloper, line 65: Ok I might be lost here... Why creating a spring message reader and then using the spring message writer? should this be just `CloudEventUtils.toReader(event).read(new MessageBuilderMessageWriter(map))`. You don't need anything else of the above stuff then, apart the `CloudEvent` declaration after the instance check

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
dsyer: Didn't work for me. Where is it done in the reader?

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
dsyer: You can try that if you want. A test will fail. If there's a way to use `CloudEventUtils` instead that's great, but I can't see it.

dsyer: We agreed OOB to drop support for structured message output. That simplifies this code.

spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java
dsyer: We agreed OOB to drop support for structured message output. That simplifies this code.

slinkydeveloper (approved):
Ok for me we're good to go, can we remove the snapshot stuff before merging?

dsyer:
I already did that, didn't I?

slinkydeveloper (approved):

PR #322, opened 2020-12-10
AMQP javadocs

slinkydeveloper:
* Aligned the amqp javadocs to the others
* Fixed external links

### DIFF
--- a/amqp/src/main/java/io/cloudevents/amqp/ProtonAmqpMessageFactory.java
+++ b/amqp/src/main/java/io/cloudevents/amqp/ProtonAmqpMessageFactory.java
@@ -29,6 +29,7 @@ import io.cloudevents.lang.Nullable;
 import io.cloudevents.rw.CloudEventRWException;
 import io.cloudevents.rw.CloudEventWriter;
 import org.apache.qpid.proton.amqp.messaging.ApplicationProperties;
+import org.apache.qpid.proton.amqp.messaging.Section;
 import org.apache.qpid.proton.message.Message;
 
 import javax.annotation.ParametersAreNonnullByDefault;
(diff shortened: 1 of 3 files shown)


slinkydeveloper:
@Alfusainey can you check this out?

Alfusainey (commented):
LGTM

amqp/src/main/java/io/cloudevents/amqp/ProtonAmqpMessageFactory.java
```diff
@@ -41,62 +44,63 @@ private ProtonAmqpMessageFactory() {
...
+    public static MessageReader createReader(final Message message) throws CloudEventRWException {
         final byte[] payload = AmqpConstants.getPayloadAsByteArray(message.getBody());
         return createReader(message.getContentType(), message.getApplicationProperties(), payload);
     }
 
     /**
-     * Creates a MessageReader using the content-type property and payload of a proton-based message.
-     * <p>
-     * This method simply calls {@link #createReader(String, ApplicationProperties, byte[])}.
+     * Creates a {@link MessageReader} using the {@code content-type} property and payload of a proton-based {@link Message}.
```
Alfusainey, line 63: @slinkydeveloper: do you think it will help the users if we mention here that this method will  create a structured reader? With the binary reader, the application-properties is mandatory but it is  optional for a structured reader

slinkydeveloper: I think we should not have at all this method TBH, since the user doesn't usually know if she/he's reading binary or structured mode events.

Alfusainey: i was thinking the same.. then how about exposing a single `createReader` API that accepts a proton `Section body` instead of `byte[]`? i.e
```Java
createReader(final String contentType, final ApplicationProperties props, @Nullable final Section payload)
```
Then the factory class can internally convert the body Section to `byte[]`, saving the user from doing so before using the API. WDYT?

slinkydeveloper: Let me try to do it

slinkydeveloper: How does it looks like now?

Alfusainey: perfect! LGTM

Alfusainey (approved):

PR #333, opened 2020-12-15
Use scheduler instead of StartupEvent. Resolves #332

### LINKED ISSUE #332: Quarkus example - nondeterministic behaviour of the startup method
In Quarkus Arc the behaviour for when the CDI StartupEvent are sent have changed between 1.7 and 1.10 (unclear to me in which version).   

This causes the quarkus rest example to fail on 1.10 since the startup method is executed before the rest endpoint is started. It would be safer to use the quarkus-scheduler like this.

```
    @Scheduled(every="2s")
    public void init() {
        Uni.createFrom().item(createEvent(userCount++)).subscribe().with( event -> {
            LOGGER.info("try to emit user: {}", event.getId());
            userClient.emit(event);
        });
    }
```

tqvarnst:


slinkydeveloper:
Hi @tqvarnst, can you sign the DCO?

examples/restful-ws-quarkus/src/main/java/io/cloudevents/examples/quarkus/client/UserEventsGenerator.java
```diff
@@ -31,16 +32,15 @@
...
-            .with(event -> {
-                LOGGER.info("try to emit user: {}", event.getId());
-                userClient.emit(event);
-            });
+    @Scheduled(every="2s")
+    public void init() {
+        Uni.createFrom().item(createEvent(userCount++)).subscribe().with( event -> {
+            LOGGER.info("try to emit user: {}", event.getId());
+            userClient.emit(event);
+        });
```
slinkydeveloper, line 43: Far more clean, I like it

slinkydeveloper (approved):

examples/restful-ws-quarkus/src/main/java/io/cloudevents/examples/quarkus/client/UserEventsGenerator.java
matejvasek: I didn't tested this, but is the `Uni<>` needed here? Wouldn't `userClient.emit(createEvent(userCount++))` work as well?

tqvarnst: Sure that works. I was trying to keep a bit of Mutiny in there, but I guess this is cleaner and examples should be clean. I will update the PR.

slinkydeveloper:
Thanks for the contribution!

PR #345, opened 2021-02-13
Fix #344 - Added links to http4k CE implementation docs.

### LINKED ISSUE #344: Question: How to showcase support from http4k
We've just added support for the CE spec in [http4k](https://http4k.org) and were wondering how we could feature this support in this SDK repo. The integration itself was pretty trivial - it was great that there wasn't much to do for us to integrate cleanly with your SDK and JSON format so sending and consuming CloudEvents is simple in a normal http4k app. You can see from the [example](https://www.http4k.org/guide/modules/cloud_events/) that we managed to hide the machinery and mostly just expose the `CloudEvent` and `CloudEventData` types.

Anyway, because it's all written in Kotlin, and we've got a [monorepo](https://github.com/http4k/http4k) for all http4k assets (with a single BOM), we like keeping the various lib integration code close to the rest - which is why we're housed it over there.

For this repo, we could add just a mention with a link to the main http4k repo example on the our site, or we can add a full example showing the integration at work in this repo. The only concern there is that the example would be Kotlin, so this might be a hurdle.

We are happy to do the work for this - just need a steer from your team as to if this is something you'd consider for this repo. 😄

daviddenton:


slinkydeveloper (approved):

slinkydeveloper:
You need to sign the dco

daviddenton:
done 😄

slinkydeveloper:
Thanks!

PR #348, opened 2021-02-19
formats: Add support for protobuf format

epaulson10:
At my company, we are interested in adopting CloudEvents, but we also are leveraging protocol buffers more and more. While the protocol-buffer format for CloudEvents has been released in the specification, it doesn't seem to have been widely adopted in any of the SDKs as of yet. During a company hackathon, I took a stab at introducing support for it, the result of which is this PR.

I'd love to get some feedback from the community.

Commit message:
------
Adds support for the Protocol Buffer CloudEvent format defined at
https://github.com/cloudevents/spec/blob/v1.0.1/protobuf-format.md.

Compiles the Proto3 file taken from the spec repo into generated Java
protobuf classes. These classes are used to convert the SDK
representation of a CloudEvent to and from the protobuf format.

Signed-off-by: Erik Paulson <epaulson@apexclearing.com>

PR #352, opened 2021-03-03
protobuf format - FOR DISCUSSION

JemDay:
Pushed as an example for discussion.

- not all test-cases have been configured.
- data handling is incomplete
- for comment

### DIFF
--- a/formats/proto/src/main/java/io/cloudevents/proto/JsonValueCloudEventData.java
+++ b/formats/proto/src/main/java/io/cloudevents/proto/JsonValueCloudEventData.java
@@ -6,12 +6,12 @@ import javax.json.JsonValue;
 
 /**
  * {@link CloudEventData} representing a {@link JsonValue}
- *
  */
 public interface JsonValueCloudEventData extends TextCloudEventData {
 
     /**
      * Obtain the {@link JsonValue} representation
+     *
      * @return The {@link JsonValue}
      */
     JsonValue getJsonValue();
--- a/formats/proto/src/main/java/io/cloudevents/proto/JsonValueCloudEventData.java
+++ b/formats/proto/src/main/java/io/cloudevents/proto/JsonValueCloudEventData.java
@@ -1,0 +1,19 @@
+package io.cloudevents.proto;
+
+import io.cloudevents.CloudEventData;
+
+import javax.json.JsonValue;
+
+/**
+ * {@link CloudEventData} representing a {@link JsonValue}
+ *
+ */
+public interface JsonValueCloudEventData extends TextCloudEventData {
+
+    /**
+     * Obtain the {@link JsonValue} representation
+     * @return The {@link JsonValue}
+     */
+    JsonValue getJsonValue();
+
+}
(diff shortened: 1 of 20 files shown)


JemDay:
I'm at bit of a loss a to why this isn't compiling ... I can execute he entire build locally ..

formats/proto/src/main/java/io/cloudevents/proto/ProtoCloudEvent.java
```diff
@@ -0,0 +1,392 @@
...
+import java.util.Base64;
+import java.util.Set;
+import java.util.function.Predicate;
+
+import static java.util.stream.Collectors.toSet;
+
+/**
+ * A CloudEvent that is backed by a Google Protobuf structure.
+ */
+class ProtoCloudEvent implements io.cloudevents.CloudEvent {
```
slinkydeveloper, line 27: I would love to avoid reimplementing `CloudEvent`, i prefer the #348 that copies values from the proto cloudevent

formats/proto/src/main/java/io/cloudevents/proto/JsonValueCloudEventData.java
```diff
@@ -0,0 +1,19 @@
+package io.cloudevents.proto;
+
+import io.cloudevents.CloudEventData;
+
+import javax.json.JsonValue;
+
+/**
+ * {@link CloudEventData} representing a {@link JsonValue}
+ */
+public interface JsonValueCloudEventData extends TextCloudEventData {
```
slinkydeveloper, line 10: This sounds interesting, worth merging with #348 i guess

formats/proto/src/test/java/io/cloudevents/proto/ProtoFormatTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+import java.io.*;
+import java.net.URL;
+import java.util.stream.Stream;
+
+import static com.google.common.truth.extensions.proto.ProtoTruth.assertThat;
+import static io.cloudevents.core.test.Data.*;
+import static org.junit.jupiter.api.Assertions.*;
+
+class ProtoFormatTest {
```
slinkydeveloper, line 21: More testing is always better :)

slinkydeveloper:
Seems like something is breaking the javadoc tool, not really sure why... Maybe some dep?

formats/proto/src/main/java/io/cloudevents/proto/JsonValueCloudEventData.java
slinkydeveloper: No wait i'm sorry, i wanted to point out to `ProtoCloudEventData`, not this one.

formats/proto/src/main/java/io/cloudevents/proto/ProtoCloudEvent.java
JemDay: I did think abut that, my concern was really from an efficiency standpoint to avoid copying values around when potentially the SKD is being used in the context of a router where maybe there's just a need to access the 'type' (for example) to select a transport/endpoint to send it to.

If the preferred model is to always clone the values I can certainly change to that for now to get stuff going but it might be worth doing some profiling to see if an optimization might be appropriate in the future.

JemDay:
Closing in favor of #348

PR #355, opened 2021-03-08
Flesh out the docs in the Spring module a bit

dsyer:


docs/spring.md
```diff
@@ -38,7 +40,147 @@ details).
...
+    return CloudEventBuilder.from(event) //
+            .withId(UUID.randomUUID().toString()) //
+            .withSource(URI.create("https://spring.io/foos")) //
+            .withType("io.spring.event.Foo") //
+            .withData(event.getData().toBytes()) //
+            .build();
+}
+```
+
+Both structured and unstructured events are supported. So if you know that the `CloudEvent` data can be converted to a `Foo` then you can also use the `CloudEventHttpUtils` to deal with HTTP headers and stick to POJOs in the handler method. Example:
```
slinkydeveloper, line 73: s/unstructured/binary

slinkydeveloper: > So if you know that the `CloudEvent` data can be converted to a `Foo`

More than this, I think you should explain that you can do that only if you're certain the input is a binary cloudevent. So maybe what about:

> So if you know that the input `CloudEvent` is in HTTP binary mode and the data can be converted to a `Foo`

docs/spring.md
```diff
@@ -38,7 +40,147 @@ details).
...
+        .withData(event.getData().toBytes())
+        .build();
+}
+```
+
+(If the application was a webapp with `spring-cloud-function-web` you would need the HTTP converters or codecs as well, per the example above.)
+
+### Generic Encoder and Decoder
+
+Some applications present Cloud Events as binary data, but do not have "headers" like in HTTP or messages. For those use cases there is a lower level construct in Spring, and this project provides implementations in the form of `CloudEventEncoder` and `CloudEventDecoder`. Since the headers are not available in the surrounding abstraction, these only support "structured" Cloud Events, where the attributes and data are packed together in the same byte buffer. As an example in an RSockets application you can register them like this:
```
slinkydeveloper, line 150: `s/"structured"/_structured_`

slinkydeveloper (commented):
Sounds good, just some minor comments

docs/spring.md
```diff
@@ -38,7 +40,147 @@ details).
...
+            .build();
+}
+```
+
+Both structured and binary events are supported. So if you know that the `CloudEvent` is in binary mode and the data can be converted to a `Foo`, then you can also use the `CloudEventHttpUtils` to deal with HTTP headers and stick to POJOs in the handler method. Example:
+
+```java
+@PostMapping("/echo")
+public ResponseEntity<Foo> echo(@RequestBody Foo foo, @RequestHeader HttpHeaders headers) {
+    CloudEvent attributes = CloudEventHttpUtils.fromHttp(headers) //
```
slinkydeveloper, line 78: Can you remove the `//` at the end of the line? Same in other examples

slinkydeveloper (commented):
Sorry I didn't saw this one in the previous PR

slinkydeveloper (approved):

PR #358, opened 2021-03-11
Specification Compliant handling of numeric context attributes

### LINKED ISSUE #357: Numeric attribute value handling..
Numeric context attributes are not being handled in a spec-conforment manner; the CE specification only support "Integer" as a valid value type and requires other numeric values to be represented as String.

The json format [specification](https://github.com/cloudevents/spec/blob/v1.0.1/json-format.md#2-attributes) further clarifies the expected behavior.

I've verified the error using the following scenario which results in the json-jackson format unit tests failing..

```
public static final CloudEvent V1_WITH_NUMERIC_EXT = CloudEventBuilder.v1()
        .withId(ID)
        .withType(TYPE)
        .withSource(SOURCE)
        .withExtension("integer", 42)
        .withExtension("decimal", new BigDecimal("42.42"))
        .withExtension("float", 4.2f)
        .withExtension("long", new Long(4200))
        .build();
```

Expected wire format..

```
{
    "specversion": "1.0",
    "id": "1",
    "type": "mock.test",
    "source": "http://localhost/source",
    "integer": 42,
    "decimal": "42.42",
    "float": "4.2",
    "long": "4200"
}
```

I will start a PR...

JemDay:
- Added tests case to verify expected handling of numeric context attributes.
- Updated serializer.

Ideally we'd change the `CloudEventContextWriter` and replace the set(name, Number) method with a set(name, Integer) method to ensure that all future implementations.

I'm happy to make that change now based on feedback.

Closes #357 

### DIFF
--- a/api/src/main/java/io/cloudevents/rw/CloudEventContextWriter.java
+++ b/api/src/main/java/io/cloudevents/rw/CloudEventContextWriter.java
@@ -83,8 +83,26 @@ public interface CloudEventContextWriter {
      * @return self
      * @throws CloudEventRWException if anything goes wrong while writing this extension.
      * @throws IllegalArgumentException if you're trying to set the specversion attribute.
+     *
+     * @deprecated CloudEvent specification only permits {@link Integer} type as a
+     * numeric value.
      */
     default CloudEventContextWriter withContextAttribute(String name, Number value) throws CloudEventRWException {
+        return withContextAttribute(name, value.toString());
+    }
+
+    /**
+     * Set attribute with type {@link Integer}.
+     * This setter should not be invoked for specversion, because the writer should
+     * already know the specversion or because it doesn't need it to correctly write the value.
+     *
+     * @param name  name of the attribute
+     * @param value value of the attribute
+     * @return self
+     * @throws CloudEventRWException if anything goes wrong while writing this extension.
+     * @throws IllegalArgumentException if you're trying to set the specversion attribute.
+     */
+    default CloudEventContextWriter withContextAttribute(String name, Integer value) throws CloudEventRWException {
         return withContextAttribute(name, value.toString());
     }
 
--- a/api/src/main/java/io/cloudevents/rw/CloudEventContextWriter.java
+++ b/api/src/main/java/io/cloudevents/rw/CloudEventContextWriter.java
@@ -85,7 +85,7 @@ public interface CloudEventContextWriter {
      * @throws IllegalArgumentException if you're trying to set the specversion attribute.
      *
      * @deprecated CloudEvent specification only permits {@link Integer} type as a
-     * numeric value.
+     * numeric val
      */
     default CloudEventContextWriter withContextAttribute(String name, Number value) throws CloudEventRWException {
         return withContextAttribute(name, value.toString());
--- a/api/src/main/java/io/cloudevents/rw/CloudEventContextWriter.java
+++ b/api/src/main/java/io/cloudevents/rw/CloudEventContextWriter.java
@@ -85,7 +85,7 @@ public interface CloudEventContextWriter {
      * @throws IllegalArgumentException if you're trying to set the specversion attribute.
      *
      * @deprecated CloudEvent specification only permits {@link Integer} type as a
-     * numeric val
+     * numeric value.
      */
     default CloudEventContextWriter withContextAttribute(String name, Number value) throws CloudEventRWException {
         return withContextAttribute(name, value.toString());
--- a/formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
+++ b/formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
@@ -134,7 +134,17 @@ class CloudEventDeserializer extends StdDeserializer<CloudEvent> {
                             writer.withContextAttribute(extensionName, extensionValue.booleanValue());
                             break;
                         case NUMBER:
-                            writer.withContextAttribute(extensionName, extensionValue.numberValue());
+
+                            final Number numericValue = extensionValue.numberValue();
+
+                            // Only 'Int' values are supported by the specification
+
+                            if (numericValue instanceof Integer){
+                                writer.withContextAttribute(extensionName, (Integer) numericValue);
+                            } else{
+                                throw CloudEventRWException.newInvalidAttributeValue(extensionName,extensionValue, null);
+                            }
+
                             break;
                         case STRING:
                             writer.withContextAttribute(extensionName, extensionValue.textValue());
--- a/formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
+++ b/formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
@@ -142,7 +142,7 @@ class CloudEventDeserializer extends StdDeserializer<CloudEvent> {
                             if (numericValue instanceof Integer){
                                 writer.withContextAttribute(extensionName, (Integer) numericValue);
                             } else{
-                                throw CloudEventRWException.newInvalidAttributeValue(extensionName,extensionValue, null);
+                                throw CloudEventRWException.newInvalidAttributeType(extensionName,numericValue);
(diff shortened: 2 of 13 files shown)


slinkydeveloper:
> Ideally we'd change the CloudEventContextWriter and replace the set(name, Number) method with a set(name, Integer) method to ensure that all future implementations.

This makes sense, how about PR'ing it deprecating the old one? We can keep `set(name, Number)` around for a while and remove it in 3/4 releases

Thanks for finding out this issue!

slinkydeveloper (commented):
I guess what we miss in this PR is to reject JSONs with fields including decimal numbers, because they are invalid per spec, right?.

We can tackle it in a followup if you want to.

JemDay:
@slinkydeveloper - I'll make the change for the context-writer as proposed.

I agree the second issue is probably best addressed as a follow-up item, I'll log an issue for that so we can agree on the approach before implementing.

formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventSerializer.java
```diff
@@ -65,15 +65,23 @@ public CloudEventContextWriter withContextAttribute(String name, String value) t
...
         @Override
-        public CloudEventContextWriter withContextAttribute(String name, Number value) throws CloudEventRWException {
+        public CloudEventContextWriter withContextAttribute(String name, Number value) throws CloudEventRWException
+        {
+            // Only Integer types are supported by the specification
+            if (value instanceof Integer) {
+                this.withContextAttribute(name, (Integer) value);
+            } else {
+                // Default to string representation for other numeric values
+                this.withContextAttribute(name, value.toString());
```
slinkydeveloper, line 75: Maybe we should fail here with a `CloudEventRWException`? Maybe this one? https://github.com/cloudevents/sdk-java/blob/master/api/src/main/java/io/cloudevents/rw/CloudEventRWException.java#L133

JemDay: Ideally we (or me!) would go through the entire SDK and remove any implementations of the deprecated function which would then result in it automagically becoming a String representation.

My worry would be if you changed this now to throw an exception it might break other stuff.

slinkydeveloper: > My worry would be if you changed this now to throw an exception it might break other stuff.

I don't think so. What is going to happen is that the sdk is becoming more strict when accepting jsons, which is fine. I prefer to increase the strictness of the input and then break the methods (which is going to happen in later sdk versions). Does it makes sense?

JemDay: Ack ... Will take a look

formats/json-jackson/src/test/java/io/cloudevents/jackson/JsonFormatTest.java
```diff
@@ -120,6 +123,29 @@ void throwExpectedOnInvalidSpecversion() {
...
+
+        try {
+
+            CloudEvent ce = getFormat().deserialize(input);
+
+            Assertions.fail("Expected Exception did not occur");
+
+        } catch (EventDeserializationException ede){
+        }
+
```
slinkydeveloper, line 146: Can you use assertj assertions for failures? https://assertj.github.io/doc/#assertj-core-exception-assertions-assertThatThrownBy

api/src/main/java/io/cloudevents/rw/CloudEventContextWriter.java
```diff
@@ -85,7 +85,7 @@ default CloudEventContextWriter withContextAttribute(String name, OffsetDateTime
      * @throws IllegalArgumentException if you're trying to set the specversion attribute.
      *
      * @deprecated CloudEvent specification only permits {@link Integer} type as a
-     * numeric value.
+     * numeric val
```
slinkydeveloper, line 88: why that change?

formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
```diff
@@ -134,7 +134,17 @@ public JsonMessage(JsonParser p, ObjectNode node) {
...
-                            writer.withContextAttribute(extensionName, extensionValue.numberValue());
+
+                            final Number numericValue = extensionValue.numberValue();
+
+                            // Only 'Int' values are supported by the specification
+
+                            if (numericValue instanceof Integer){
+                                writer.withContextAttribute(extensionName, (Integer) numericValue);
+                            } else{
+                                throw CloudEventRWException.newInvalidAttributeValue(extensionName,extensionValue, null);
```
slinkydeveloper, line 145: Should it be https://github.com/cloudevents/sdk-java/blob/master/api/src/main/java/io/cloudevents/rw/CloudEventRWException.java#L133?

JemDay: doh... probably..

JemDay: Actually, having thought a bit more, maybe the one I chose is more appropriate in this situation as it will actually cause the problematic value to be logged along with the attribute name. The other exception only reports the problematic attribute name.

I'm now leaning to leaving it as-is ... thoughts ?!?

slinkydeveloper: For me you should use invalid type as exception kind... Maybe you need another constructor for it that takes and logs the value?

slinkydeveloper: You can have an overload like:

```
    public static CloudEventRWException newInvalidAttributeType(String attributeName, Object value) {
        return new CloudEventRWException(
            CloudEventRWExceptionKind.INVALID_ATTRIBUTE_TYPE,
            "Invalid attribute/extension type for \"" + attributeName + "\". Type: " + value.getClass().getCanonicalName() + ". Value: " + value
        );
    }
```

formats/json-jackson/src/test/java/io/cloudevents/jackson/JsonFormatTest.java
JemDay: Ack

api/src/main/java/io/cloudevents/rw/CloudEventContextWriter.java
JemDay: Finger trouble I suspect - fixing

formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
JemDay: Ack .. will do

formats/json-jackson/src/test/java/io/cloudevents/jackson/JsonFormatTest.java
```diff
@@ -120,6 +122,22 @@ void throwExpectedOnInvalidSpecversion() {
...
+    @MethodSource("badJsonContent")
+    /**
+     * JSON content that should fail deserialization
+     * as it represents content that is not CE
+     * specification compliant.
+     */
+    void verifyDeserializeError(String inputFile){
+
+        byte[] input = loadFile(inputFile);
+        boolean exceptionThrown = false;
```
slinkydeveloper, line 135: Is that still useful?

slinkydeveloper (commented):
Just one small nit and we're good to go!

formats/json-jackson/src/test/java/io/cloudevents/jackson/JsonFormatTest.java
JemDay: doh - good catch..

slinkydeveloper (approved):

slinkydeveloper:
Thanks for the PR! Can we also consider #360 done?

PR #362, opened 2021-03-25
Throw exceptions when attempting to handle unsupported CloudEvent formats

### LINKED ISSUE #361: Handling of unsupported Content-Types
Occurs for JAX-RS, probably affects other protocols..

When payload is received that claims to contain structured content (eg is has an application/cloudevent+xxx content-type) that doesn't match one of the registered formats an attempt is made to look at the headers to  see if the content is a binary cloud event.

This causes issues in situations if the producer sends a structured format that the receiver doesn't have registered, especially in situations where context-attributes are also present in protocol header as this can fool the SDK into thinking the content is in binary-mode - which is a false assumption.

I will open a PR ...

JemDay:
- Modfy logic that selects between structured and binary modes during reception.
- Introduced new test scenarios to veify behavior.

Closes #361 

### DIFF
--- a/core/src/main/java/io/cloudevents/core/message/impl/MessageUtils.java
+++ b/core/src/main/java/io/cloudevents/core/message/impl/MessageUtils.java
@@ -59,6 +59,15 @@ public class MessageUtils {
             EventFormat format = EventFormatProvider.getInstance().resolveFormat(ct);
             if (format != null) {
                 return structuredMessageFactory.apply(format);
+            } else {
+                /**
+                 * The format wasn't one we support, but if it's part of the
+                 * CloudEvent family it indicates it's a structured
+                 * representation that we can't interpret.
+                 */
+                if (ct.startsWith("application/cloudevents")) {
+                    throw newUnknownEncodingException();
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/cloudevents/core/message/impl/MessageUtils.java
```diff
@@ -59,6 +59,15 @@ public static MessageReader parseStructuredOrBinaryMessage(
             EventFormat format = EventFormatProvider.getInstance().resolveFormat(ct);
             if (format != null) {
                 return structuredMessageFactory.apply(format);
+            } else {
+                /**
+                 * The format wasn't one we support, but if it's part of the
+                 * CloudEvent family it indicates it's a structured
+                 * representation that we can't interpret.
+                 */
+                if (ct.startsWith("application/cloudevents")) {
```
slinkydeveloper, line 68: I see what you're trying to do here. AFAIK this logic is nowhere in the cloudevents sdks. But I see it's following this wording: https://github.com/cloudevents/spec/blob/v1.0.1/http-protocol-binding.md#3-http-message-mapping

slinkydeveloper (approved):

core/src/main/java/io/cloudevents/core/message/impl/MessageUtils.java
slinkydeveloper: It might be worth updating the other sdks too

PR #367, opened 2021-04-01
Avoid allocating an array on extension validation

pierDipi:
As per title.
There is no need to allocate another array.


slinkydeveloper:
Sounds good, can you rebase?

slinkydeveloper (approved):

PR #376, opened 2021-04-28
 CloudEvent.toString() does not include null attributes

### LINKED ISSUE #326: CloudEvent.toString() should not include null attributes
I have only checked io.cloudevents.core.v1.CloudEventV1 but rather than concatenation null values for optional attributes they should be omitted.  I am open to fixing this and doing a pull request if this is preferred.. potentially its already fixed in a branch?  Thank you everyone for the work.

slinkydeveloper:
Fix #326


matzew (approved):
/lgtm
/approve

PR #392, opened 2021-05-17
[CESQL] Constant folding

slinkydeveloper:
This PR includes:

* A new visitor to visit the `ExpressionInternal` tree and mutate it, in order to execute some manipulations (like the constant folding)
* Introduced `BaseUnaryExpression` to simplify a bit the visitor thing
* Implemented `ConstantFoldingExpressionVisitor` to perform the constant folding
* Moved `ExpressionInternal` in expressions package
* Added some tests covering both the tck with constant folding enabled and with constant folding disabled

In the next PRs I'll reorganize a bit the code in more packages, in order to avoid the mess we have now

### DIFF
--- a/sql/src/main/java/io/cloudevents/sql/impl/ParserImpl.java
+++ b/sql/src/main/java/io/cloudevents/sql/impl/ParserImpl.java
@@ -18,21 +18,20 @@ import java.util.List;
 
 public class ParserImpl implements Parser {
 
-    private final boolean constantFolding;
-
     private static class SingletonContainer {
         private final static ParserImpl INSTANCE = new ParserImpl(true);
     }
 
     /**
-     * @return instance of {@link ParserImpl}
+     * @return default instance of {@link ParserImpl}
      */
     public static Parser getInstance() {
         return ParserImpl.SingletonContainer.INSTANCE;
     }
 
-    // TODO this is super bad, remove it!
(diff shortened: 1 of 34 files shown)


slinkydeveloper:
To simplify the review of the PR, I can split in several PRs

sql/src/main/java/io/cloudevents/sql/impl/ParserImpl.java
```diff
@@ -27,7 +31,9 @@ public static Parser getInstance() {
         return ParserImpl.SingletonContainer.INSTANCE;
     }
 
-    public ParserImpl() {
+    // TODO this is super bad, remove it!
+    public ParserImpl(boolean constantFolding) {
```
pierDipi, line 35: I agree! :)

pierDipi (approved):
Great stuff!

LGTM

sql/src/main/java/io/cloudevents/sql/impl/ParserImpl.java
slinkydeveloper: yeah I wanted to create some builder like thing

slinkydeveloper: Can you check it out now? I've added a ParserBuilder

PR #396, opened 2021-06-04
[CESQL] Simpler interface to perform casting

slinkydeveloper:
Provide a simpler interface to perform casting outside an expression. This is useful when the output result of the expression needs to be coerced to a specific type, eg when using for filtering, you always need to coerce the result back to boolean.


slinkydeveloper:
cc @pierDipi

pierDipi (approved):

PR #415, opened 2021-08-31
fix NPE in CloudEventDeserializer when deserializing with optional header with null value

bdshadow:
Signed-off-by: Dmitrii Bocharov <dmitrii.bocharov@embedit.cz>


pierDipi:
Thanks @bdshadow!

PR #429, opened 2021-12-07
Ignore not valid extension name in jackson CloudEventDeserializer

### LINKED ISSUE #428: Ignore not valid extension key format in CloudEventDeserializer (jackson)
CloudEventDeserializer by cloudevents-json-jackson throws an exception when an invalid extension key is included.
Since the extension attribute is not a required value, why don't CloudEventDeserializer ignore invalid attributes and run Deserialize?

I changed to ignore the invalid extensionName in CloudEventDeserializer.

mhyeon-lee:
CloudEventDeserializer by cloudevents-json-jackson throws an exception when an invalid extension key is included.
Since the extension attribute is not a madatory value, why don't CloudEventDeserializer ignore invalid attributes and run Deserialize?

I modified to ignore the invalid extensionName in CloudEventDeserializer. 

Fixes #428

### DIFF
--- a/formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
+++ b/formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
@@ -126,7 +126,13 @@ class CloudEventDeserializer extends StdDeserializer<CloudEvent> {
 
                 // Now let's process the extensions
                 node.fields().forEachRemaining(entry -> {
-                    String extensionName = entry.getKey();
+                    String extensionName = entry.getKey().toLowerCase();
+
+                    // ignore not valid extension name
+                    if (!this.isValidExtensionName(extensionName)) {
+                        return;
+                    }
+
                     JsonNode extensionValue = entry.getValue();
 
                     switch (extensionValue.getNodeType()) {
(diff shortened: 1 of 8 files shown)


formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
```diff
@@ -126,7 +126,13 @@ public JsonMessage(JsonParser p, ObjectNode node) {
 
                 // Now let's process the extensions
                 node.fields().forEachRemaining(entry -> {
-                    String extensionName = entry.getKey();
+                    String extensionName = entry.getKey().toLowerCase();
+
+                    // ignore not valid extension name
+                    if (!this.isValidExtensionName(extensionName)) {
+                        return;
+                    }
```
pierDipi, line 134: This might be surprising or unwanted.

Can we make it an opt-in behavior?

pierDipi (changes requested):
Thanks for sending this PR!

formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
mhyeon-lee: @pierDipi 
Thank you for your feedback
I added two options for jackson CloudEventDeserializer
https://github.com/cloudevents/sdk-java/pull/429/commits/cc4963e0cd6a8e83a30529cf2a9ceecc41f93e48

formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormat.java
```diff
@@ -47,41 +47,95 @@
...
+     * @see #withForceExtensionNameLowerCaseDeserialization()
+     * @see #withForceIgnoreInvalidExtensionNameDeserialization()
      */
-    public JsonFormat(boolean forceDataBase64Serialization, boolean forceStringSerialization) {
+    public JsonFormat(
+        boolean forceDataBase64Serialization,
+        boolean forceStringSerialization,
+        boolean forceExtensionNameLowerCaseDeserialization,
+        boolean forceIgnoreInvalidExtensionNameDeserialization
+    ) {
```
pierDipi, line 70: Can we create a class that holds these configurations `forceDataBase64Serialization`, `forceStringSerialization`, etc rather than having a lot of boolean parameters?

Also, if we create a class we can create a different constructor so that we're not making breaking changes.

pierDipi: Maybe we can have a builder class with a fluent interface to build such configuration objects.

mhyeon-lee: @pierDipi 
Create JsonFormatOptions class with builder.
It has serialization / deserialization boolean options.
And it does not breaking change

pierDipi (approved):
Thanks!

LGTM

PR #445, opened 2022-02-07
Optimize `isCloudEventsHeader` check

pierDipi:
Instead of using:
```java
key.substring(0, CE_PREFIX.length()).toLowerCase().startsWith(CE_PREFIX);
```
that allocates 2 new strings, we can just use:
```java
key.regionMatches(true /* ignoreCase */, 0, CE_PREFIX, 0, CE_PREFIX.length());
```

Signed-off-by: Pierangelo Di Pilato <pierdipi@redhat.com>

### DIFF
--- a/core/src/main/java/io/cloudevents/core/impl/StringUtils.java
+++ b/core/src/main/java/io/cloudevents/core/impl/StringUtils.java
@@ -21,6 +21,10 @@ import javax.annotation.Nonnull;
 
 final public class StringUtils {
 
+    private StringUtils() {
+        // Prevent construction.
+    }
+
     public static boolean startsWithIgnoreCase(@Nonnull final String s, @Nonnull final String prefix) {
         return s.regionMatches(true /* ignoreCase */, 0, prefix, 0, prefix.length());
     }
--- a/core/src/main/java/io/cloudevents/core/impl/StringUtils.java
+++ b/core/src/main/java/io/cloudevents/core/impl/StringUtils.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2018-Present The CloudEvents Authors
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.cloudevents.core.impl;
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/cloudevents/core/impl/StringUtils.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ *
+ */
+
+package io.cloudevents.core.impl;
+
+import javax.annotation.Nonnull;
+
+final public class StringUtils {
+
```
matzew, line 23: private ctor?

matzew (dismissed):
/lgtm
/approve

core/src/main/java/io/cloudevents/core/impl/StringUtils.java
pierDipi: Done

kafka/src/main/java/io/cloudevents/kafka/impl/KafkaBinaryMessageReaderImpl.java
```diff
@@ -44,7 +44,7 @@ protected boolean isContentTypeHeader(String key) {
 
     @Override
     protected boolean isCloudEventsHeader(String key) {
-        return key.length() > 3 && key.substring(0, KafkaHeaders.CE_PREFIX.length()).startsWith(KafkaHeaders.CE_PREFIX);
+        return key.length() > KafkaHeaders.CE_PREFIX.length() && key.startsWith(KafkaHeaders.CE_PREFIX);
```
JemDay, line 47: Any particular reason this isn't using your new method ?

Also ... Might be worth adding the null-check on `key` (most of the other readers do this)

JemDay (commented):
While you're here .. is there value in adding a generic implementation of _isCloudEventsHeader(prefix, key)_.

Then all of these implementations can be simplified to use that method and remove all the duplicated code

kafka/src/main/java/io/cloudevents/kafka/impl/KafkaBinaryMessageReaderImpl.java
pierDipi: > Any particular reason this isn't using your new method ?

it wasn't case-insensitive before like other ones and I think the reason is that other methods are HTTP based (headers normalization) vs this one being Kafka based. 

Do you know if the spec is case-insensitive for every defined protocol binding?

Honestly, from a user perspective, I would be surprised if a header isn't considered a CE header due to the case, so I agree with you.

matzew (approved):
/lgtm
/approve

kafka/src/main/java/io/cloudevents/kafka/impl/KafkaBinaryMessageReaderImpl.java
pierDipi: Hey @JemDay since most of the comments aren't strictly related to the original goal of the PR and the code is strictly equivalent to the existing one, I think, we can follow up on these comments in a different PR/issue.

PR #469, opened 2022-08-19
Restful-ws jakarta ee9 namespace support

### LINKED ISSUE #468: Jakarta EE9 and onwards support for http-restful-ws package
The current Jakarta http-restful-ws package uses the old `javax.*` namespace for the jax-rs packages that were replaced in Jakarta EE9 with `jakarta.*`. As such J EE9 and the upcoming EE10 are currently incompatible with the http-restful-ws package.

Both Resteasy and Jersey latest versions are JEE9 compatible, though Spring are planning Jee9 support 4Q2022, but for support for boot 3 and framework version 6, Jee9 support will be required

abutch3r:
Fixes #468 

Add support for newer  Jakarta EE versions

This is just a way to show what changes are needed to support the move from the `javax` -> `jakarta` namespaces which is required for the latest version of Microprofile and the upcoming Spring-boot 3 and Spring-framework 6. which both support Jakarta EE9 and also with the upcoming JEE10 release.

The PR works by copying the contents of the original restful-ws definitions into a new module and during the copy replaces javax with jakarata. the Pom also depends on a newer jakarta ws-api package that uses the jakarta named packages. This ensures a single code base between the two packages with minimal build overhead

This includes a port of the reasteasy test case as the Jersey scenario requires Java11 and given Spring don't support it yet, that has not been ported either.

The name of the package is just a place holder and more then happy to change to meet the requirements of the project. 

* I am currently working on a microprofile integration test scenario, but this does include an example of Microprofile with OpenLiberty using the new package


http/restful-ws-jakarta/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+            <layout>default</layout>
+            <releases>
+                <enabled>true</enabled>
+            </releases>
+        </pluginRepository>
+    </pluginRepositories>
+
+    <properties>
+        <jakarta-ee.version>3.0.0</jakarta-ee.version>
+        <jersey.version>2.30.1</jersey.version>
```
rvowles, line 45: this needs to be 3.0.x - jersey 2 still uses javax and to support jakarta only, it should change to version 3

rvowles:
I would also like to see this merged as jakarta has been the new standard for around 2 years now.

http/restful-ws-jakarta/pom.xml
abutch3r: @rvowles thanks for the spot. Have updated the version to latest 3.0.x version and ported the Jersey test case over to the new version

http/restful-ws-jakarta/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+                                <delete dir="src/main"/>
+                                <copy todir="src">
+                                    <fileset dir="../restful-ws\src" />
+                                    <filterchain>
+                                        <replaceregex pattern="javax." replace="jakarta." />
+                                    </filterchain>
+                                </copy>
+                                <!--<replace dir="src/main/java/io/cloudevents/http/restful/ws" token="javax" value="jakarta">
+                                    <include name=" **/*.java" />
+                                </replace>-->
```
pierDipi, line 84: Can we remove this commented code?

pierDipi (commented):
Hi @abutch3r, thanks for the PR.

I did a quick scan of the PR, one concern that I have is that the new module `http/restful-ws-jakarta-integration-tests` used for testing different implementations will be published as part of the release even though it is only used for testing.
Can we avoid publishing the new module `http/restful-ws-jakarta-integration-tests` since it is not meant to be used by users?

http/restful-ws-jakarta/pom.xml
abutch3r: @pierDipi I am working on a new Integration test case at the moment and as part of that commit I will address both items

abutch3r: @pierDipi I have added the new test case, found a few other items that needed cleaning up and adding documentation for the new package. so I consider this ready for a deeper review

docs/core.md
```diff
@@ -110,6 +110,6 @@ CloudEvent extensions can be materialized in their respective POJOs using the
 import io.cloudevents.core.extensions.DistributedTracingExtension;
 import io.cloudevents.core.provider.ExtensionProvider;
 
-DistributedTracingExtension dte = ExtensionsParser.getInstance()
+DistributedTracingExtension dte = ExtensionProvider.getInstance()
```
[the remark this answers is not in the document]
abutch3r: @pierDipi a rebase with associated Merge commit seems to of done the trick, not sure why these came out as they did as it was me rebasing back to upstream/master that pulled in those changes

pierDipi (approved):
LGTM

@rvowles do you have any other comments?

rvowles:
No @pierDipi looks solid to me!

pierDipi:
Thanks @abutch3r and @rvowles!

Issue #470, opened 2022-08-23
JSON content type detection not spec compliant


henning-cg:
The logic for detecting if the datacontenttype is JSON seems to be contrary to the current spec.

The spec states that:

> Such a content type is defined as one having a [media subtype](https://tools.ietf.org/html/rfc2045#section-5) equal to json or ending with a +json format extension. That is, a datacontenttype declares JSON-formatted content if its media type, when stripped of parameters, has the form `*/json` or `*/*+json`. If the datacontenttype is unspecified, processing SHOULD proceed as if the datacontenttype had been specified explicitly as application/json

but the code at https://github.com/cloudevents/sdk-java/blob/91251365307dcd11ffe8558cfb52397af3852a1b/formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormat.java#L217 only checks for empty datacontenttype, application/json or text/json, and is not taking into account for example the extension case "application/something+json"

PR #471, opened 2022-08-24
support dynamic data content type

segator:
Fixes https://github.com/cloudevents/sdk-java/issues/470
Now checks if dataContentType matches the next regex.

`^(application|text)\/([a-zA-Z]+\+)?json$")`

This regex support 
`application/foobar+json`
or standard
```
application/json
text/json
```

### DIFF
--- a/formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormat.java
+++ b/formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormat.java
@@ -47,7 +47,7 @@ public final class JsonFormat implements EventFormat {
     /**
      * Suppoted Content type
      */
-    private static final Pattern PATTERN_CONTENT_TYPE = Pattern.compile("^(application|text)\\/([a-zA-Z]+\\+)?json$");
+    private static final Pattern JSON_CONTENT_TYPE = Pattern.compile("^(application|text)\\/([a-zA-Z]+\\+)?json$");
     private final ObjectMapper mapper;
     private final JsonFormatOptions options;
 
@@ -218,6 +218,6 @@ public final class JsonFormat implements EventFormat {
 
     static boolean dataIsJsonContentType(String contentType) {
         // If content type, spec states that we should assume is json
-        return contentType == null || PATTERN_CONTENT_TYPE.matcher(contentType).matches();
+        return contentType == null || JSON_CONTENT_TYPE.matcher(contentType).matches();
     }
 }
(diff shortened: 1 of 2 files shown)


n3wscott:
Makes sense. Likely a change we should do in the go sdk too to support `application/foobar+json`

matejvasek:
Just maybe rename `PATTERN_CONTENT_TYPE` to `JSON_CONTENT_TYPE`.

segator:
> Just maybe rename `PATTERN_CONTENT_TYPE` to `JSON_CONTENT_TYPE_PATTERN`.

updated

formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormat.java
```diff
@@ -43,7 +44,10 @@ public final class JsonFormat implements EventFormat {
      * Content type associated with the JSON event format
      */
     public static final String CONTENT_TYPE = "application/cloudevents+json";
-
+    /**
+     * Suppoted Content type
```
matejvasek, line 48: I am not sure if it's good to call it "Supported Content type".
Other content type should work too, they will just be exposed as `byte[]` not objects IIRC.

segator: Understood, any suggestion for the naming?

JemDay: How about something along the lines of "_JSON Data Content Type Discriminator"_ - long-winded I known ;-)

JemDay:
While we're changing this test-case could we also enhance it to ensure that the 'datacontenttype' in the deserialized event matches the 'contentType' that was used to construct the original event (from your test argument list)

Just an extra level of sanity checking.

formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormat.java
segator: @JemDay  Done!

segator:
> While we're changing this test-case could we also enhance it to ensure that the 'datacontenttype' in the deserialized event matches the 'contentType' that was used to construct the original event (from your test argument list)
> 
> Just an extra level of sanity checking.

Done!

JemDay:
Thanks LGTM....

pierDipi (approved):
Thanks @segator, LGTM

Issue #413, opened 2021-08-19
NATS Java client for NATS sender, publisher, receiver and also for NATS JetStream integration


srimadhavan:
I am working on to Integrate the Cloud Events with NATS Event Driven Architecture. I am seeing there is no JAVA client library for NATS and NATS JetStream. Please add the Java client for sender, receiver, publish and subscribe for NATS and NATS Jetstream from Cloud Event client.

pierDipi:
Contributions are welcome!

rvowles:
I'm working on this.

rvowles:
The bit thats taking the time is the header value encoding as per the spec here: https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/nats-protocol-binding.md - its not URL encoding and its not anything standard and is going to require a fair degree of work.

lilytrotter:
Is anyone working on this? Seems it stalled on the previous attempt at the PR stage. I'd be interested if not

I'm trying to encourage Cloudevents adoption at work, having NATS integration would smooth the path. I have studied the Golang implementation and the spec so I have a fair sense of it.

rvowles:
I already implemented and published it under my own company public GitHub repo because I couldn't get a review completed after several months of waiting. It has been successfully used in several releases of our open source feature flagging platform.

https://github.com/featurehub-io/featurehub-cloudevents

rvowles:
@lilytrotter did you try the one I linked?

PR #476, opened 2022-09-13
NATS bindings

rvowles:
This addresses issue #413 - I understand it will take a fair amount to review and there may be plenty of feedback. I  tend to squash commits to retain only a single commit in a PR so please let me know if this is not desirable.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -27,7 +27,7 @@ Using the Java SDK you can:
 ## Supported features
 
 |                                                    | [v0.3](https://github.com/cloudevents/spec/tree/v0.3) | [v1.0](https://github.com/cloudevents/spec/tree/v1.0) |
-| :------------------------------------------------: | :---------------------------------------------------: | :---------------------------------------------------: |
+|:--------------------------------------------------:| :---------------------------------------------------: | :---------------------------------------------------: |
 |                  CloudEvents Core                  |                  :heavy_check_mark:                   |                  :heavy_check_mark:                   |
 |               AMQP Protocol Binding                |                          :x:                          |                          :x:                          |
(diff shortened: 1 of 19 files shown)


rvowles:
Alternatively it might be more worthwhile for my opensource project (FeatureHub) to maintain this binding outside the lifecycle of the sdk-java and just have the sdk-java point to it, as we are active users of NATS.

rvowles:
In the meantime as I need this, I have released these bindings under FeatureHub, along with AWS Kinesis and Google PubSub bindings. https://github.com/featurehub-io/featurehub-cloudevents

pierDipi:
Hi @rvowles, thanks for the PR!

I'm slowly reviewing the PR, it might take a while since it is big.

rvowles:
....?

JemDay:
A non-functional change to consider, @pierDipi I'd like your comment here as well.

While I've been working offline on some MQTT bindings what stuck me was that the way our code is organized wasn't very consistent, by that I mean all the different format implementations live under <root>/formats but everything else just floats  under the root.

My proposal is to start housing the different bindings under a /bindings folder and subdivide that by transport and then by implementation.

In my MQTT work I've got this structure..

<root>/bindings/mqtt/paho
<root>/bindings/mqtt/hivemq

Can we consider moving this under bindings/nats ??

JemDay:
Another proposal ... 

I'm a massive fan of encapsulation and 'implementation hiding' ... by placing all of the implementing classes in their own package they have essentially become publicly accessible.

Given that the 'MessageFactory' represents the programatic interface to the module that should be the only thing implementors need to worry about and see documentation about IMHO.

As such my proposal would be to move all of the 'impl' classes into the main package and change their visibility to package-private.

rvowles:
Given there seems no intention to integrate this MR, and I have already released it elsewhere (along with Kinesis and Google PubSub), I will close this.

PR #477, opened 2022-09-14
Set correct url to naming conventions.

gklijs:
I noticed the url in the Java doc was not correct. As it turns out using a regex is about ten times slower for this case, so I removed that part of the commit.

### DIFF
--- a/core/src/main/java/io/cloudevents/core/impl/BaseCloudEventBuilder.java
+++ b/core/src/main/java/io/cloudevents/core/impl/BaseCloudEventBuilder.java
@@ -213,7 +213,7 @@ public abstract class BaseCloudEventBuilder<SELF extends BaseCloudEventBuilder<S
      *
      * @param name the extension name
      * @return true if extension name is valid, false otherwise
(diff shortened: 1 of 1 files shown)


pierDipi:
>  also think it would be preferable to use a regex for validating the extension name.

Would you mind expanding on the why?

gklijs:
It might be faster, although, with the expected small size of an extension name, that might not be the case. It does limit the lines of code, but the complexity of a regex might outweigh that.

pierDipi:
> It might be faster, although, with the expected small size of an extension name, that might not be the case. It does limit the lines of code, but the complexity of a regex might outweigh that.

Exactly, a proper benchmark should be added to validate the change isn't a perf regression for various cases.

pierDipi:
Hi @gklijs, thanks for the fix and the quick update, can you please fix the DCO check as explained here: https://github.com/cloudevents/sdk-java/pull/477/checks?check_run_id=8348631721?

gklijs:
Should be good now.

Issue #481, opened 2022-09-23
2.4.0 breaks JSON Deserialization


fredoboulo:
Hi,

We can't update io.cloudevents:cloudevents-json-jackson:2.3.0 to 2.4.0.

Our cloudEvents have a **datacontenttype** with value **application/json;charset=utf-8**, which leads to JSON deserialization error.
The [newly introduced Pattern in JsonFormat](https://github.com/cloudevents/sdk-java/blob/master/formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormat.java#L50) does not take parameters into account.

Reading the spec, it seems a bit confusing

- [Payload Serialization](https://github.com/cloudevents/sdk-java/blob/master/formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormat.java#L50) handles the parameters stripping
- [Payload Deserialization](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/json-format.md#312-payload-deserialization) is silent about parameters.

Would it be possible to enrich the pattern to take the parameters into account?

pierDipi:
Hi @fredoboulo, thanks for reporting!

I think it makes sense what you're suggesting.

cc @JemDay @segator

pierDipi:
@fredoboulo since nobody raised any concern so far, can you contribute this change to the SDK?

segator:
Sorry for the delay guys I were on vacations with limited access to mail.
@pierDipi  so the idea is to support in the regex attributes like the charset one

`foo/barr;attr1=value&attr2=value... `?

PR #484, opened 2022-10-03
fix: strip parameters from data content types to assess if it's JSON format

fredoboulo:
Fixes https://github.com/cloudevents/sdk-java/issues/481

Signed-off-by: Frederic Delechamp <fdelechamp@guidewire.com>

### DIFF
--- a/formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormat.java
+++ b/formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormat.java
@@ -47,7 +47,7 @@ public final class JsonFormat implements EventFormat {
     /**
      * JSON Data Content Type Discriminator
      */
-    private static final Pattern JSON_CONTENT_TYPE_PATTERN = Pattern.compile("^(application|text)\\/([a-zA-Z]+\\+)?json$");
+    private static final Pattern JSON_CONTENT_TYPE_PATTERN = Pattern.compile("^(application|text)\\/([a-zA-Z]+\\+)?json(;.*)*$");
     private final ObjectMapper mapper;
     private final JsonFormatOptions options;
 
--- a/formats/json-jackson/src/test/java/io/cloudevents/jackson/JsonFormatTest.java
+++ b/formats/json-jackson/src/test/java/io/cloudevents/jackson/JsonFormatTest.java
@@ -48,6 +48,22 @@ class JsonFormatTest {
     private final ObjectMapper mapper = new ObjectMapper();
 
     @ParameterizedTest
+    @MethodSource("jsonContentTypes")
+    void isJsonContentType(String contentType) {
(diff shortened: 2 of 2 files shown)


fredoboulo:
Fixes https://github.com/cloudevents/sdk-java/issues/481

JemDay:
** This _may_ apply to much more than just the JSON format serialization.

Don't the various transport bindings also check for content-type such as application/cloudevents+json ?? theoretically those checks should also include this type of stripping.

I'm not sure if now is the time but we may want to break all of this stuff out into a some core utilities that can be shared across all of the formats and transports .. most specifically "take a content-type and strip off all the trailing attributes" and "isJson(stripped-content-type)" type functions.

Not sure if we should address this now or add it as a followup .. it might need a bit more analysis to see where everybody is doing a comparison on content-type.

fredoboulo:
@JemDay I'm not sure this comment is directed to me but...

it makes sense to me, for the potential to be needed elsewhere. However, I'm not familiar with this particular project, so I'd individually prefer to defer refactoring in a different context than this PR.

formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormat.java
```diff
@@ -47,7 +47,7 @@ public final class JsonFormat implements EventFormat {
     /**
      * JSON Data Content Type Discriminator
      */
-    private static final Pattern JSON_CONTENT_TYPE_PATTERN = Pattern.compile("^(application|text)\\/([a-zA-Z]+\\+)?json$");
+    private static final Pattern JSON_CONTENT_TYPE_PATTERN = Pattern.compile("^(application|text)\\/([a-zA-Z]+\\+)?json(;[\\s\\w\"!#$%&'*+.^_`{|}~=-]*)*$");
```
JemDay, line 50: I wonder if this regex can be simplified so that it simply matches anything following the ";" so something like `"(;.*)*$"` - my regex may be a bit rusty. ;-)

JemDay (dismissed):
LGTM .. simplifying the regex would be nice for efficient processing but could be a follow-on activity.

formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormat.java
fredoboulo: Thanks, I was overdoing it :)

pierDipi (approved):
LGTM

pierDipi:
Thanks @fredoboulo!

pierDipi:
@fredoboulo would you be willing to backport the merged commit https://github.com/cloudevents/sdk-java/commit/6362bfbcd67f41bc3b9fbb977796e0ebeafc80c0 to the 2.4 branch for a patch release?

fredoboulo:
Hi @pierDipi, that's about cherry-picking and making a PR in branch 2.4?

PR #491, opened 2022-12-08
Use UTF-8 when using getBytes

### LINKED ISSUE #488: [security] Reliance on default encoding
# Description
Multiple instances were identified in which the getByte() standard Java API is used without specifying any encoding. Doing so causes the Java SDK to rely on the system default encoding, which can differ across platforms and systems used by event actors and cause unexpected differences in processing of event data.
The specification states that appropriate and RFC-compliant encodings MUST be followed, but the implementation in the Java SDK and documentation should be improved to highlight the importance of matching encoding across actors.
Not all observed instances are necessarily problematic, as they are handling binary data. However, this behavior should be documented and handled in the SDK implementation, documentation, and supplied examples.
```
import io.cloudevents.CloudEvent;
import io.cloudevents.core.builder.CloudEventBuilder;

import java.net.URI;

final CloudEvent event = CloudEventBuilder.v1()
    .withId("000")
    .withType("example.demo")
    .withSource(URI.create("http://example.com"))
    .withData("text/plain","Hello world!".getBytes())
    .build();
```
[code](https://github.com/cloudevents/sdk-java/blob/b9eaa2fcaaf5569552e39ece4fce4a99064145e9/docs/core.md#L28-L40)

```
private byte[] getBinaryData(Message<?> message) {
     Object payload = message.getPayload();
     if (payload instanceof byte[]) {
       return (byte[]) payload;
     }
     else if (payload instanceof String) {
            return ((String) payload).getBytes(Charset.defaultCharset());
     }
     return null;
}
```
[code](https://github.com/cloudevents/sdk-java/blob/b9eaa2fcaaf5569552e39ece4fce4a99064145e9/spring/src/main/java/io/cloudevents/spring/messaging/CloudEventMessageConverter.java#L93-L102)

# Exploit Scenario
The event producer, the intermediary (using the SDK), and the consumer use different default encodings for their systems. Without acknowledging a fixed encoding, the data is handled and processed using an unintended encoding, resulting in unexpected behavior.

# Recommendations
Short term, improve the SDK documentation to highlight the importance of matching encoding acros actors.
Long term, review all similar instances across the SDK and improve test cases to cover handling of message and data encoding.

# References
- [The Java Tutorials; Byte Encodings and Strings](https://docs.oracle.com/javase/tutorial/i18n/text/string.html)
- [PMD New rule: Reliance on default charset #2186](https://github.com/pmd/pmd/issues/2186)

duglin:
Closes #488

Signed-off-by: Doug Davis <dug@microsoft.com>

### DIFF
--- a/core/src/test/java/io/cloudevents/core/data/PojoCloudEventDataTest.java
+++ b/core/src/test/java/io/cloudevents/core/data/PojoCloudEventDataTest.java
@@ -1,5 +1,7 @@
 package io.cloudevents.core.data;
 
+import java.nio.charset.StandardCharsets;
+
 import org.junit.jupiter.api.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -8,7 +10,7 @@ class PojoCloudEventDataTest {
 
     @Test
     void testWrapAndMemoization() {
-        PojoCloudEventData<Integer> data = PojoCloudEventData.wrap(10, i -> i.toString().getBytes());
+        PojoCloudEventData<Integer> data = PojoCloudEventData.wrap(10, i -> i.toString().getBytes(StandardCharsets.UTF_8));
 
         assertThat(data.getValue())
             .isEqualTo(10);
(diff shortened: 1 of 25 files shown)


lance (commented):
/lgtm

pierDipi (approved):

PR #529, opened 2023-03-03
[2.4] Fix javadoc error

pierDipi:


pierDipi:
Let's see how the build goes with this PR ...

PR #535, opened 2023-03-06
Make ProtoCloudEventData consistent with spec.

### LINKED ISSUE #532: [protobuf] 'create' and 'read' aren't entirely symmetrical
> It's starting to come back to me now ... with respect to the data handling there's no guarantee that a receiving application/intermediary will be able to hydrate the actual ProtoBuf message from the data on reception, that's why the 'Any'  construct is returned so that the business app can do the appropriate type mapping if needed, or pass it on elsewhere if required.
>
> Granted this does mean the 'create' and 'read' aren't entirely symmetrical ..

_Originally posted by @JemDay in https://github.com/cloudevents/sdk-java/issues/523#issuecomment-1440827380_

JemDay:
Modified ProtoCloudEventData to  _always_ return a Protobuf _Any_ object - this ensures it is coherent with the Protobuff CloudEvent format specification.

It remains possible to wrap any Protobuf 'Message' object directly (which includes an 'Any')  as a convienience to reduce application code.

Closes #532

### DIFF
--- a/docs/protobuf.md
+++ b/docs/protobuf.md
@@ -11,20 +11,22 @@ This module provides the Protocol Buffer (protobuf) `EventFormat` implementation
 Protobuf runtime and classes generated from the CloudEvents
 [proto spec](https://github.com/cloudevents/spec/blob/v1.0.1/spec.proto).
 
+# Setup
 For Maven based projects, use the following dependency:
 
 ```xml
 <dependency>
     <groupId>io.cloudevents</groupId>
     <artifactId>cloudevents-protobuf</artifactId>
-    <version>2.3.0</version>
+    <version>x.y.z</version>
 </dependency>
 ```
 
+No further configuration is required is use the module.
+
 ## Using the Protobuf Event Format
 
-You don't need to perform any operation to configure the module, more than
-adding the dependency to your project:
+### Event serialization
 
 ```java
 import io.cloudevents.CloudEvent;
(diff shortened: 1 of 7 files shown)


pierDipi (commented):
Do we have or can we add tests on the new getAny when data is  `binary_data` or `text_data` [1] ?

[1] https://github.com/cloudevents/spec/blob/3877083f8396cfb01b7b3e8adf1738f248af3aff/cloudevents/formats/cloudevents.proto#L38-L40

docs/protobuf.md
```diff
@@ -44,5 +46,51 @@ byte[]serialized = EventFormatProvider
...
+
+// Deserialize the event.
+CloudEvent myEvent = eventFormat.deserialize(raw);
+
+// Get the Data
+CloudEventData eventData = myEvent.getData();
+
+if (ceData instanceOf ProtoCloudEventData) {
+
+    // Obtain the protobuff 'any'
```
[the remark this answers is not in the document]
JemDay: Ack

JemDay:
> Do we have or can we add tests on the new getAny when data is `binary_data` or `text_data` [1] ?
> 

Yes .. in two situations:

- The 'serialize' test verifies that CE's are being serialized correctly .. that set includes both binary and text data.
- The original 'roundTripTest' does a  exp-wire-format -> CloudEvent -> act-wire-format round-trip to ensure (actual == expected) at the wire level - that test set also includes both binary and textual cloud event data

formats/protobuf/src/main/java/io/cloudevents/protobuf/ProtoDataWrapper.java
```diff
@@ -53,17 +60,22 @@ public boolean equals(Object obj) {
...
 
-        final Any lhsAny = getAsAny(this.getMessage());
-        final Any rhsAny = getAsAny(rhs.getMessage());
+        if (lhsAny == rhsAny) {
+            return true;
+        }
+
+        if ((lhsAny == null) || (rhsAny == null)) {
+            return false;
+        }
```
pierDipi, line 78: lhsAny or rhsAny cannot be null since the constructor has:
```
        Objects.requireNonNull(protoMessage);

        if (protoMessage instanceof Any) {
            protoAny = (Any) protoMessage;
        } else {
            protoAny = Any.pack(protoMessage);
        }
```

```suggestion
```

formats/protobuf/src/main/java/io/cloudevents/protobuf/ProtoSerializer.java
```diff
@@ -244,16 +246,14 @@ public CloudEvent end(CloudEventData data) throws CloudEventRWException {
...
-                        if (m instanceof Any) {
-                            protoBuilder.setProtoData((Any) m);
-                        }else {
-                            protoBuilder.setProtoData(Any.pack(m));
-                        }
+                    final Any anAny = protoData.getAny();
+
+                    if (anAny != null) {
+                        protoBuilder.setProtoData(anAny);
                     }
```
pierDipi, line 255: `anAny` cannot be null
```suggestion
                        protoBuilder.setProtoData(protoData.getAny());
```

pierDipi: However, in this class might be acceptable to keep the check since technically the 2 are decoupled

formats/protobuf/src/main/java/io/cloudevents/protobuf/ProtoDataWrapper.java
JemDay: ;-) ... thanks

formats/protobuf/src/main/java/io/cloudevents/protobuf/ProtoSerializer.java
JemDay: Modified to throw an exception if no data is present.

pierDipi (approved):
Thanks @JemDay !

PR #557, opened 2023-04-28
2.4

robp94:


robp94:
@pierDipi ok lets if it builds. What do you want to check else if it is backward compatible? I looked over the changes and there seem to be no real changes, only additions. But I am not too familiar with the project.

pierDipi:
Thanks for the PR @robp94, I believe it's better to release this in 2.5, I've just created a 2.5 branch that we can use to push this backport.

Let's see how the build goes.

Issue #560, opened 2023-04-29
Optimize BaseCloudEvent#readExtensions method


mxsm:
![image](https://user-images.githubusercontent.com/15797831/235286990-f119f38e-32ac-42cf-aa8b-8aca0fb1ce01.png)
Optimize BaseCloudEvent#readExtensions method

PR #561, opened 2023-04-29
[ISSUE #560]Optimize BaseCloudEvent#readExtensions method

mxsm:
Optimize BaseCloudEvent#readExtensions method

### DIFF
--- a/core/src/main/java/io/cloudevents/core/impl/BaseCloudEvent.java
+++ b/core/src/main/java/io/cloudevents/core/impl/BaseCloudEvent.java
@@ -65,20 +65,14 @@ public abstract class BaseCloudEvent implements CloudEvent, CloudEventReader, Cl
     }
 
     protected void readExtensions(CloudEventContextWriter writer) throws CloudEventRWException {
-        // TODO to be improved
         for (Map.Entry<String, Object> entry : this.extensions.entrySet()) {
-            if (entry.getValue() instanceof String) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/cloudevents/core/impl/BaseCloudEvent.java
```diff
@@ -65,20 +65,14 @@ public <T extends CloudEventWriter<V>, V> V read(CloudEventWriterFactory<T, V> w
...
-                writer.withContextAttribute(entry.getKey(), (Boolean) entry.getValue());
-            } else if (entry.getValue() instanceof URI) {
-                writer.withContextAttribute(entry.getKey(), (URI) entry.getValue());
-            } else if (entry.getValue() instanceof OffsetDateTime) {
-                writer.withContextAttribute(entry.getKey(), (OffsetDateTime) entry.getValue());
-            } else if (entry.getValue() instanceof byte[]) {
-                writer.withContextAttribute(entry.getKey(), (byte[]) entry.getValue());
+            Object value = entry.getValue();
+            if (value instanceof String || value instanceof Number || value instanceof Boolean || value instanceof URI) {
+                writer.withContextAttribute(entry.getKey(), value.toString());
```
pierDipi, line 71: Using the `.toString()` is not equivalent to previous implementation that was using typed methods.

The original issue is not clear what the optimization is?

PR #578, opened 2023-06-06
Avro Compact Format

alexec:


formats/avro-compact/src/test/java/io/cloudevents/avro/compact/AvroCompactFormatTest.java
```diff
@@ -0,0 +1,71 @@
...
+                .withData("", BytesCloudEventData.wrap(new byte[0]))
+                .build();
+
+        byte[] serialized = format.serialize(event);
+
+        assertNotNull(serialized);
+
+        CloudEvent deserialized = format.deserialize(serialized);
+
+        assertEquals(event, deserialized);
```
pierDipi, line 68: Is it possible to verify that `deserialized` after `format.serialize(event)` is byte to byte equal to `serialized` ?

alexec:
Wow. I’ve used DCO on many projects now and it is make-work junk.

PR #604, opened 2024-01-31
feat: JSON format to assume JSON content-type where possible

matejvasek:
For JSON structured CE we may assume JSON `datacontenttype` as long as `data` is present int the CE.

### DIFF
--- a/formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
+++ b/formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
@@ -42,18 +42,21 @@ import java.io.IOException;
 class CloudEventDeserializer extends StdDeserializer<CloudEvent> {
     private final boolean forceExtensionNameLowerCaseDeserialization;
     private final boolean forceIgnoreInvalidExtensionNameDeserialization;
+    private final boolean disableDataContentTypeDefaulting;
 
     protected CloudEventDeserializer() {
-        this(false, false);
+        this(false, false, false);
     }
 
     protected CloudEventDeserializer(
         boolean forceExtensionNameLowerCaseDeserialization,
-        boolean forceIgnoreInvalidExtensionNameDeserialization
+        boolean forceIgnoreInvalidExtensionNameDeserialization,
+        boolean disableDataContentTypeDefaulting
     ) {
         super(CloudEvent.class);
         this.forceExtensionNameLowerCaseDeserialization = forceExtensionNameLowerCaseDeserialization;
         this.forceIgnoreInvalidExtensionNameDeserialization = forceIgnoreInvalidExtensionNameDeserialization;
+        this.disableDataContentTypeDefaulting = disableDataContentTypeDefaulting;
     }
 
     private static class JsonMessage implements CloudEventReader {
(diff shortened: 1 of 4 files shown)


formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
```diff
@@ -92,6 +92,9 @@ public <T extends CloudEventWriter<V>, V> V read(CloudEventWriterFactory<T, V> w
 
                 // Parse datacontenttype if any
                 String contentType = getOptionalStringNode(this.node, this.p, "datacontenttype");
+                if (contentType == null && this.node.has("data")) {
```
matzew, line 95: I think this is aligned w/ the `JSON-format.md` file, section 3.1:

https://github.com/cloudevents/spec/blob/main/cloudevents/formats/json-format.md#31-handling-of-data

As for the non-binary / none base64 case it states:
> If the `datacontenttype` is unspecified, processing SHOULD proceed as if the
`datacontenttype` had been specified explicitly as `application/json`.

However the spec says _should_ not _must_.

Perhaps that's part of the issue?

ricardozanini: Yes, it can introduce dubious questioning as the one we are having in the Knative community. @duglin can you chime in?

matejvasek:
> Should this be done behind a configurable flag as it is a "should" vs "must" in the spec? and perhaps as opt-out?

That may be a good idea. How would we call it? `forceDatacontenttypeDefaulting` `forceDatacontenttypeJson` ?

matejvasek:
PTAL @pierDipi @matzew

formats/json-jackson/src/main/java/io/cloudevents/jackson/CloudEventDeserializer.java
matejvasek: ping @duglin

duglin: I'm not sure I'm understanding the question. Is it: why is it `SHOULD` and not a `MUST` in that part of the spec?  If so, I can't remember for sure, but I suspect it was done to allow for some out of band knowledge built into the event producer that allowed for it to "just know" what the real type is and take the appropriate serialization action. But w/o any extra knowledge things should assume/default to JSON.  Make any sense?

formats/json-jackson/src/main/java/io/cloudevents/jackson/JsonFormatOptions.java
```diff
@@ -61,11 +64,14 @@ public boolean isForceIgnoreInvalidExtensionNameDeserialization() {
...
     }
 
+    public boolean isDataContentTypeDefaultingDisabled() { return this.disableDataContentTypeDefaulting; }
+
     public static class JsonFormatOptionsBuilder {
         private boolean forceDataBase64Serialization = false;
         private boolean forceStringSerialization = false;
         private boolean forceExtensionNameLowerCaseDeserialization = false;
         private boolean forceIgnoreInvalidExtensionNameDeserialization = false;
+        private boolean disableDataContentTypeDefaulting =false;
```
pierDipi, line 74:
```suggestion
        private boolean disableDataContentTypeDefaulting = false;
```

pierDipi:
One small style nit

matejvasek:
> One small style nit

fixed @pierDipi

pierDipi (approved):
Thanks @matejvasek !

LGTM

Issue #638, opened 2024-05-30
Error assembling JAR: Invalid automatic module name: ''


Jummartinezro:
Im' running with the following setup:

```bash
PS sdk-java> java -version
openjdk version "11.0.23" 2024-04-16 LTS
OpenJDK Runtime Environment Microsoft-9394293 (build 11.0.23+9-LTS)
OpenJDK 64-Bit Server VM Microsoft-9394293 (build 11.0.23+9-LTS, mixed mode, sharing)
PS sdk-java> mvn --version
Apache Maven 3.9.6 (bc0240f3c744dd6b6ec2920b3cd08dcc295161ae)
Maven home: C:\Program Files\apache-maven-3.9.6
Java version: 11.0.23, vendor: Microsoft, runtime: C:\Program Files\Microsoft\jdk-11.0.23.9-hotspot
Default locale: fr_FR, platform encoding: Cp1252
OS name: "windows 11", version: "10.0", arch: "amd64", family: "windows"
PS sdk-java>
```

While doing a `mvn package`, I have the following error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-jar-plugin:3.3.0:jar (default-jar) on project cloudevents-http-restful-ws-integration-tests-common: Error assembling JAR: Invalid automatic module name: '' -> [Help 1]
```

The same happens when I'm trying to package other modules (Like `cloudevents-kafka-example`)

pierDipi:
Fixed by https://github.com/cloudevents/sdk-java/pull/639

PR #639, opened 2024-05-30
fix: invalid automatic module name

Jummartinezro:
Inspired by [Automatic-Module-Name: Calling All Java Library Maintainers](https://dzone.com/articles/automatic-module-name-calling-all-java-library-maintainers), I added the module names for the non-specified modules on their respective `pom.xml` file

Solves #638


pierDipi (approved):
Thanks!

PR #603, opened 2024-01-30
Sync CESQL tck tests

Cali0707:
I recently realized that the tck tests for CESQL are not up to date with the new ones added in the spec repo. This PR just copies those changes over into sdk-go, and registers the new test file.

cc @duglin @pierDipi

PR #641, opened 2024-06-19
Cesql v1 fixes

### LINKED ISSUE #618: Port tck tests once the CESQL spec stabilizes more
Recently there have been many changes to the CESQL spec. Once that stabilizes a bit, we should port this to this SDK and implement any fixes/changes required.

cc @pierDipi I would be happy to work on this if assigned

Cali0707:
Update the CESQL implementation to pass the v1 spec conformance tests.

The main change in this PR is that now the `EvaluationResult` type is used internally as well as for the final return value of the expressions, instead of using a thrower interface to collect errors over time. This allows for handling of specific exceptions that are encountered in subexpressions (for example, missing attributes). Outside of that, there are various small fixes to get the SDK to match the v1 spec.

I _tried_ to organize my work by commit, but since I wanted to get everything to pass before I did that I only sort of managed to do that. So, the commits are generally organized by what changed in them, but there may be some extra stuff in each commit. Sorry about that!

cc @pierDipi @matzew 
Fixes #618


sql/src/main/java/io/cloudevents/sql/FunctionSignature.java
```diff
@@ -18,6 +18,8 @@ public interface FunctionSignature {
      */
     Type typeOfParameter(int i) throws IllegalArgumentException;
 
+    Type returnType();
```
pierDipi, line 21:
```suggestion
    /**
     * @return function return type
     */
    Type returnType();
```

pierDipi: Just for consistency with other functions

sql/src/main/java/io/cloudevents/sql/impl/ExceptionFactoryImpl.java
```diff
@@ -0,0 +1,131 @@
...
+                EvaluationException.ErrorKind.CAST,
+                interval,
+                expression,
+                "Cannot cast " + from + " to " + to + ": " + cause.getMessage(),
+                cause
+            );
+
+            if (this.shouldFail) {
+                throw exception;
+            }
```
pierDipi, line 49: maybe "shouldThrow" is a better name ?

sql/src/main/java/io/cloudevents/sql/impl/ExceptionFactoryImpl.java
```diff
@@ -0,0 +1,131 @@
...
+            return exception;
+        };
+    }
+
+    public EvaluationException missingAttribute(Interval interval, String expression, String key) {
+        final EvaluationException exception = new EvaluationException(
+            EvaluationException.ErrorKind.MISSING_ATTRIBUTE,
+            interval,
+            expression,
+            "Missing attribute " + key + " in the input event. Perhaps you should check with 'EXISTS " + key + "' if the input contains the provided key?",
```
pierDipi, line 59: That's an ideal error message with the suggested fix

sql/src/main/java/io/cloudevents/sql/impl/ExceptionFactoryImpl.java
```diff
@@ -0,0 +1,131 @@
...
+            interval,
+            expression,
+            "Division by zero: " + dividend + " / 0",
+            null
+        );
+
+        if (this.shouldFail) {
+            throw exception;
+        }
+        return exception;
```
pierDipi, line 113: This could become:

```suggestion
        return mathError(interval, expression, "Division by zero: " + dividend + " / 0");
```

sql/src/main/java/io/cloudevents/sql/impl/runtime/EvaluationResult.java
```diff
@@ -2,20 +2,79 @@
...
+        this(value, exception, null, null);
+    }
+
+    public EvaluationResult(Object value) {
+        this.value = value;
+        this.exceptions = new ArrayList<>();
+        this.latestException = null;
+    }
+
+    public EvaluationResult wrap(EvaluationResult other) {
```
pierDipi, line 41: Should we call this "wrapExceptions" as I believe it would make callers code a little more explicit on what are we wrapping

pierDipi: including the overloaded one

sql/src/test/java/io/cloudevents/sql/TCKTestSuite.java
pierDipi, line 1: Can we fix the test suite? https://github.com/cloudevents/sdk-java/pull/603#pullrequestreview-1853465105

Cali0707: Should be fixed now!

pierDipi:
> Error: [ERROR] error: Error fetching URL: https://qpid.apache.org/releases/qpid-proton-j-0.33.7/api/ (java.io.FileNotFoundException: https://qpid.apache.org/releases/qpid-proton-j-0.33.7/api/package-list)

Can you completely remove that link?

Cali0707:
Done - can you recheck this @pierDipi

pierDipi:
Thanks @Cali0707

pierDipi (approved):
LGTM

PR #648, opened 2024-07-01
Allow bumping versions in `/docs` only

pierDipi:
This will allow changing the version in the documentation website when a new release is published.


pierDipi:
cc @Cali0707 @matzew

PR #650, opened 2024-07-02
Make CloudEventValidatorProvider thread safe

### LINKED ISSUE #643: Cannot invoke "java.lang.ClassLoader.getParent()" because "this.currentLoader" is null
Hi,
Since `cloudevents-core:3.0.0`, we are experiencing this error 
```java
java.lang.NullPointerException: Cannot invoke "java.lang.ClassLoader.getParent()" because "this.currentLoader" is null
	at java.base/java.util.ServiceLoader$ModuleServicesLookupIterator.hasNext(ServiceLoader.java:1075)
	at java.base/java.util.ServiceLoader$2.hasNext(ServiceLoader.java:1309)
	at java.base/java.util.ServiceLoader$3.hasNext(ServiceLoader.java:1393)
	at io.cloudevents.core.provider.CloudEventValidatorProvider.validate(CloudEventValidatorProvider.java:48)
	at io.cloudevents.core.v1.CloudEventBuilder.build(CloudEventBuilder.java:127)
```
This does not happen every time we use the builder. I suspect a race condition as `CloudEventValidatorProvider` is a Singleton that relies on the ServiceLoader class which is not thread safe according to its documentation. 

> Instances of this class are not safe for use by multiple concurrent threads.

pierDipi:
Fixes #643

### DIFF
--- a/core/src/main/java/io/cloudevents/core/provider/CloudEventValidatorProvider.java
+++ b/core/src/main/java/io/cloudevents/core/provider/CloudEventValidatorProvider.java
@@ -19,8 +19,8 @@ package io.cloudevents.core.provider;
 import io.cloudevents.CloudEvent;
 import io.cloudevents.core.validator.CloudEventValidator;
 
-import java.util.Iterator;
-import java.util.ServiceConfigurationError;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.ServiceLoader;
 
 /**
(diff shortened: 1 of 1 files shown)


pierDipi:
@flebastard can you take a look?

pierDipi:
cc @touchkey

flebastard (approved):

flebastard:
> @flebastard can you take a look?

Hi, LGTM as the `ServiceLoader.load()` is called only once when the singleton is created

PR #679, opened 2024-11-04
fix: disallow empty 'subject' header

### LINKED ISSUE #676: Java SDK allows empty subject during deserialization
Hi,

[Specification](https://github.com/cloudevents/spec/blob/v1.0/spec.json) says that "subject" is optional but if present, must be non-empty.

Doing a simple test with structured or binary event delivery shows that this is not validated by the standard deserializers - the screenshot below shows how a test expecting a 400 Bad Request actually passes with 200 OK.
image

pnt:
Specification clearly states that 'subject' is optional but if present, MUST be non-empty (spec at https://github.com/cloudevents/spec/blob/v1.0/spec.md#subject)

Fixes #676


pierDipi (approved):
Thanks, LGTM

Issue #694, opened 2025-06-18
Upgrade Protobuf version


SleimanJneidi:
Protobuf version is now on 4.x which is not compatible with protobuf 3.x, the wire format has not changed, just the java sdk. 
So existing messages should parse fine with protobuf 4.x Java SDK.

Spring Boot Grpc 0.8.0 is on 4.x already.

PR #695, opened 2025-06-18
upgrade to protobuf sdk 4.31.1

SleimanJneidi:
Upgrade protobuf to 4.3.1. 
This is the latest version of protobuf, no changes for the wire format but the library is not compatible with 3.x so we might need a major release for it

https://github.com/cloudevents/sdk-java/issues/694


pierDipi (approved):
LGTM

PR #705, opened 2026-01-08
#704: Bumped to SpringBoot4, SpringFramework7 and JDK17

Stromner:
Resolves issue #704

PR #729, opened 2026-04-24
Bumped to SpringBoot4, SpringFramework7 and JDK17 - same as #705, but with signed-ff commits

jacekbilski:
The content is the same as in #705, see the discussion there. The only two changes are: the commits are now properly signed-off and I've rebased it onto main.


lehphyro:
Hey guys, are you planning to merge this soon? It's blocking Spring Boot upgrades because of Jackson  3 breaking changes.

jacekbilski:
I wish I could say more. I believe that @stromner-stralfors and I have made all the necessary changes, the code works for me™. Anything beyond that is in the hands of the maintainers.

lbenedetto:
@duglin Can you answer if this is something you (the maintainers, not specifically you) will be working on?

duglin:
@lbenedetto let me poke some folks

duglin:
Seems reasonable to me, however, with such a large number of files changed (and some being more than just a version number change), let's wait for @Cali0707 to verify that nothing seems off.

Cali0707:
yes currently taking a look here (started yesterday, but lots of changes...)

Cali0707 (commented):
Hey @jacekbilski thanks for all the much needed hard work to modernize this SDK!

I think to make these changes more releasable we need to split this up into a couple more incremental PRs/releases:

1. Let's mark `http/restful-ws`'s public classes as `@Deprecated(forRemoval = true)`. As part of this, we should copy the code into `http/restful-ws-jakarta` and permanently switch the `javax` to `jakarta` namespacing. We do not want to remove the `restful-ws-jakarta` support in 5.0 (and can't remove either currently as neither module is actually marked as deprecated, despite the commit comment saying they are).
2. Make a 4.1.0 release after step 1 is merged
3. Bump to Java 17 baseline and apply the language modernization changes (version bump to 5.0.0-SNAPSHOT)
4. Drop the `restful-ws` module and update `restful-ws-jakarta` to Jakarta REST 4.0
5. Apply the `HttpHeaders` defensive copy fix and the `@Nullable` annotation switch (these are both needed for Spring Boot 4 forward compatibility), but keep compiling and testing against Spring Boot 3.x. This way users on both 3.x and 4.x can use the module. I thinking moving from the current 2.x all the way to 4.x in one go is too aggressive and will make it so some users can not use this.

I don't think we need to bump to Jackson 3 currently, as there is still a large portion of projects on Jackson 2. Additionally, let's stick with JUnit 5 for now to help keep the scope of changes smaller.

If you are able to start working on pieces of this in separate PRs, ping me and I'm happy to try to quickly merge them and move this process along!

jacekbilski:
Hello @Cali0707,

I'll start working on those things soon.

One note, current 4.x version is working perfectly with Spring Boot 3, I'm using it that way since over a year. So the jump to Spring Boot 4 is not that big.

lehphyro:
Hey @Cali0707, Spring Boot 4 upgrades Jackson which renames packages to `tools.jackson`, not upgrading it means users will be forced to use the deprecated compatibility module `spring-boot-jackson2`, see https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide#upgrading-jackson

Cali0707:
@jacekbilski @lehphyro I appreciate the points you are raising. My strong preference for not bumping these all now is because we have lots of different usages in different projects with different sets of dependencies.

Let's:
1. Do a release with java 17, and the different fixes (as I mentioned above), using jackson 2. We can indicate that this will be the last release on jackson 2.
2. We can do a follow up release with spring 4 and jackson 3. This should hopefully help the different consumers of this repo with time to migrate (and again, even if things are working, jumping from 2.x to 4.x is a very large jump - we really should only go one major version at a time)

jacekbilski:
@Cali0707 As we're anyway providing Jackson support in the form of `json-jackson` module, maybe providing two, one for Jackson 2 and one for Jackson 3, would be an option? I cannot tell just yet how difficult or time consuming would that be, I'll be looking into this. But this way we could provide something for everyone.

Cali0707:
@jacekbilski we could definitely do a dual package approach for jackson 2/3 (and IMO deprecate the jackson 2 package to indicate to people to move off it). I think this is a great approach

As with everything else here I'm fully on board with modernizing the dependencies/packages in this SDK, I just want to make sure that we are upgrading in a way where it is easy for all the SDK consumers to incrementally update, and where we are communicating what is being deprecated in releases before we remove support.

Cali0707:
@jacekbilski are you okay with the plan we've discussed here? Mind opening issues for the different steps so that we can close out this PR and start tackling/releasing bits of this incrementally?

jacekbilski:
@Cali0707 Fine by me. I'll open new issues and close this PR once I'll start working on it.

core/src/main/java/io/cloudevents/core/impl/CloudEventContextReaderAdapter.java
```diff
@@ -54,16 +54,16 @@ public void readAttributes(CloudEventContextWriter writer) throws RuntimeExcepti
     public void readExtensions(CloudEventContextWriter writer) throws RuntimeException {
         for (String key : event.getExtensionNames()) {
```
HattoriHenzo, line 55: Even thought this variable is design to be a key in a map, I suggest to rename it as` extensionName` or `extension` or `name`

core/src/main/java/io/cloudevents/core/impl/CloudEventContextReaderAdapter.java
```diff
@@ -54,16 +54,16 @@ public void readAttributes(CloudEventContextWriter writer) throws RuntimeExcepti
...
-                writer.withContextAttribute(key, (String) value);
-            } else if (value instanceof Number) {
-                writer.withContextAttribute(key, (Number) value);
-            } else if (value instanceof Boolean) {
-                writer.withContextAttribute(key, (Boolean) value);
-            } else if (value instanceof URI) {
-                writer.withContextAttribute(key, (URI) value);
-            } else if (value instanceof OffsetDateTime) {
-                writer.withContextAttribute(key, (OffsetDateTime) value);
+            if (value instanceof String string) {
```
HattoriHenzo, line 57: This code can be refactor to Pattern Matching with a `switch`.  (I saw that those kind of changes has already been mentioned by @Cali0707 here: Bump to Java 17 baseline and apply the language modernization changes)

core/src/main/java/io/cloudevents/core/impl/CloudEventContextReaderAdapter.java
```diff
@@ -54,16 +54,16 @@ public void readAttributes(CloudEventContextWriter writer) throws RuntimeExcepti
...
+                writer.withContextAttribute(key, string);
+            } else if (value instanceof Number number) {
+                writer.withContextAttribute(key, number);
+            } else if (value instanceof Boolean boolean1) {
```
HattoriHenzo, line 61: Rename this `boolean1` variable to `bool`

core/src/main/java/io/cloudevents/core/impl/CloudEventContextReaderAdapter.java
```diff
@@ -54,16 +54,16 @@ public void readAttributes(CloudEventContextWriter writer) throws RuntimeExcepti
...
+                writer.withContextAttribute(key, boolean1);
+            } else if (value instanceof URI rI) {
```
HattoriHenzo, line 63: Rename `rI` to `uri` maybe...

HattoriHenzo (commented):
@jacekbilski I appreciate the effort and an incremental approach is better. Are you using OpenRewrite?

jacekbilski:
@Cali0707, @duglin, I just created #732. Please take a look and if possible do a release.

My next step will be to bump Java version to 17 and do a general cleanup of the code. I'll also upgrade some major dependencies by one step, also then when the new version is already EOL. That would mean things like Spring Boot 2 -> 3, Quarkus 1 -> 2 or Jetty 9 -> 10 (or maybe even 12 as this one provides support for anything starting from Java EE 8). I'll try to keep the project and the dependencies compatible with Jakarta EE 9.

Would you be OK with that? I think that was the idea. But jumping onto even newer versions, like Spring Boot 4 or Quarkus 3, might again require a major version change on our side. The differences between Jakarta EE versions above 8 are not that drastic, but keeping the compatibility between all those projects might still be challenging.

jacekbilski:
> @jacekbilski I appreciate the effort and an incremental approach is better. Are you using OpenRewrite?

@HattoriHenzo yes, as you have correctly guessed based on those variable named, I am. I'll be including your proposals once I get that far.

jacekbilski:
@Cali0707, @duglin, my second PR is ready, see #735. Hopefully, I didn't get too much ahead and it could still be merged and released as version 4.1.1.

Next up is really jumping onto Java 17, dropping deprecated Jakarta EE 8 and moving the baseline to Jakarta EE 9. Or maybe we should jump directly to Jakarta EE 10? Like I mentioned already, the differences are not as drastic between 9, 10 and 11 as between 8 and 9. As we'll anyway be bumping major version, maybe we could/should take now a bit bigger leap? For example Quarkus has no version compatible with Jakarta EE 9, they jumped directly from 8 to 10. How do you see it?

jacekbilski:
@Cali0707, @duglin, my third PR is ready, see #739.

What's not included yet, but planned for the next PR is:
* bumping the baseline to Jakarta EE 11,
* bumping Spring Boot to version 4 (matches Jakarta EE 11, but also requires/suggests further changes),
* dropping FindBugs/JSR305 in favour of JSpecify (Spring Boot 3 is not using it yet, other projects here also not, will probably require some work),
* adding a separate module for Jackson 3 and keeping the current one.

Like you've mentioned, it's too early to delete Jackson 2 support. On the other hand, Spring Boot 4 is already using Jackson 3 while keeps compatibility with Jackson 2, but not for long, see https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-7.0-Release-Notes#jackson-3x-support. So we probably need both.

Given the changes planned, we might again require a major version jump, but I'll leave the decision to you.

I hope you're OK with the scope of current and the next PRs.

PR #732, opened 2026-05-19
Deprecating to Jakarta EE 8

jacekbilski:
As discussed in #705 and #729, see https://github.com/cloudevents/sdk-java/pull/729#pullrequestreview-4253594404, this is the first PR. The goal is to deprecate Jakarta EE 8 module and promote the Jakarta EE 9+ module.

I added Jakarta EE 9+ links in a few places where they were missing. I also fixed the links where they were still pointing to the Jakarta EE 8 module.

The examples are still using the now deprecated `cloudevents-http-restful-ws` module. The reason is - I cannot upgrade Quarkus or Jetty simply because next major releases, that are Jakarta EE 9+-compatible would also require Java at least 11, but we're still on 8.

### DIFF
--- a/http/restful-ws-jakarta/README.md
+++ b/http/restful-ws-jakarta/README.md
@@ -1,4 +1,4 @@
-# HTTP Protocol Binding for Jakarta EE9+ - Jakarta RESTful Web Services
+# HTTP Protocol Binding for Jakarta EE 9+ - Jakarta RESTful Web Services
 
 Javadocs: [![Javadocs](http://www.javadoc.io/badge/io.cloudevents/cloudevents-http-restful-ws-jakarta.svg?color=green)](http://www.javadoc.io/doc/io.cloudevents/cloudevents-http-restful-ws-jakarta)
 
--- a/http/restful-ws-jakarta/README.md
+++ b/http/restful-ws-jakarta/README.md
@@ -3,5 +3,3 @@
 Javadocs: [![Javadocs](http://www.javadoc.io/badge/io.cloudevents/cloudevents-http-restful-ws-jakarta.svg?color=green)](http://www.javadoc.io/doc/io.cloudevents/cloudevents-http-restful-ws-jakarta)
 
 Documentation: https://cloudevents.github.io/sdk-java/http-jakarta-restful-ws-jakarta
-
-The code for this package lies within the [restful-ws](https://github.com/cloudevents/sdk-java/tree/main/http/restful-ws) directory and is copied within here and has `javax.` replaced with `jakarta.`
--- a/http/restful-ws-jakarta/README.md
+++ b/http/restful-ws-jakarta/README.md
@@ -1,7 +1,7 @@
 # HTTP Protocol Binding for Jakarta EE9+ - Jakarta RESTful Web Services
 
-Javadocs: [![Javadocs](http://www.javadoc.io/badge/io.cloudevents/cloudevents-http-restful-ws.svg?color=green)](http://www.javadoc.io/doc/io.cloudevents/cloudevents-http-restful-ws)
+Javadocs: [![Javadocs](http://www.javadoc.io/badge/io.cloudevents/cloudevents-http-restful-ws-jakarta.svg?color=green)](http://www.javadoc.io/doc/io.cloudevents/cloudevents-http-restful-ws-jakarta)
 
-Documentation: https://cloudevents.github.io/sdk-java/http-jakarta-restful-ws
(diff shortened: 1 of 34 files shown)


jacekbilski:
@Cali0707 My second PR is also almost ready. I did a lot of cleaning up there, Maven wrapper and plugins, documentation. I also upgraded things to their latest Jakarta EE 8-compatible versions, just to ease upgrading them to their Jakarta EE 9-compatible versions. It's just, more or less, bringing things in sync. As such, those changes might as well belong in here. Based on your [proposed versioning](https://github.com/cloudevents/sdk-java/pull/729#pullrequestreview-4253594404), this could be version 4.1.1.

Please take a look at https://github.com/jacekbilski/sdk-java/pull/1. I could extend this PR with those changes if it would make sense to you.

Because of that I also changed the title of this PR.

I also started working on the next PR. The goal there would be moving to Java 17 and upgrading everything to Jakarta EE 9-compatible versions (Spring Boot 3, Quarkus 2, etc.).

http/restful-ws-jakarta/README.md
```diff
@@ -1,7 +1,7 @@
-# HTTP Protocol Binding for Jakarta EE9+ - Jakarta RESTful Web Services
+# HTTP Protocol Binding for Jakarta EE 9+ - Jakarta RESTful Web Services
 
-Javadocs: [![Javadocs](http://www.javadoc.io/badge/io.cloudevents/cloudevents-http-restful-ws.svg?color=green)](http://www.javadoc.io/doc/io.cloudevents/cloudevents-http-restful-ws)
+Javadocs: [![Javadocs](http://www.javadoc.io/badge/io.cloudevents/cloudevents-http-restful-ws-jakarta.svg?color=green)](http://www.javadoc.io/doc/io.cloudevents/cloudevents-http-restful-ws-jakarta)
 
-Documentation: https://cloudevents.github.io/sdk-java/http-jakarta-restful-ws
+Documentation: https://cloudevents.github.io/sdk-java/http-jakarta-restful-ws-jakarta
 
 The code for this package lies within the [restful-ws](https://github.com/cloudevents/sdk-java/tree/main/http/restful-ws) directory and is copied within here and has `javax.` replaced with `jakarta.`
```
Cali0707: @jacekbilski mind updating this bit? It is now outdated with your changes

http/restful-ws-jakarta/src/main/resources/META-INF/services/jakarta.ws.rs.ext.Providers
```diff
@@ -0,0 +1 @@
+io.cloudevents.http.restful.ws.CloudEventsProvider
```
Cali0707: @jacekbilski this is missing a license header

Cali0707 (commented):
Thanks for tackling this @jacekbilski !

Two small nits otherwise I am happy to go forwards with this change

Cali0707:
@jacekbilski regarding your other PR, let's do those as a separate follow up.

Additionally it looks like Kafka/Vertx have major version bumps there? Since those are exposed types in the public API that would be breaking changes so we should hold off on upgrading those two until we do the 5.0 set of changes. Otherwise that looks great and can be a 4.1.1

Thanks again for tackling all of this!

http/restful-ws-jakarta/README.md
jacekbilski: Fixed, I had it removed on at least one machine on which I was working on this.

http/restful-ws-jakarta/src/main/resources/META-INF/services/jakarta.ws.rs.ext.Providers
jacekbilski: I now deleted the file. It was saying that `io.cloudevents.http.restful.ws.CloudEventsProvider` implements `jakarta.ws.rs.ext.Providers`. But it does not, it never did.

jacekbilski:
> @jacekbilski regarding your other PR, let's do those as a separate follow up.

OK, as soon as this part is released I'll create a new PR.

> Additionally it looks like Kafka/Vertx have major version bumps there? Since those are exposed types in the public API that would be breaking changes so we should hold off on upgrading those two until we do the 5.0 set of changes. Otherwise that looks great and can be a 4.1.1

OK, I'll do the upgrades later.

> Thanks again for tackling all of this!

I always wanted to somehow contribute to open source, I finally got the chance.

Cali0707 (approved):
LGTM, thanks @jacekbilski !

jacekbilski:
Cool, tomorrow I'll deploy to production an app with an upgraded dependency.

On with the next PR...

PR #735, opened 2026-05-31
Cleanups before Java 17

jacekbilski:
This is the second PR of the series towards Java 17 / Spring Boot 4 / Quarkus 3 / Jakarta EE 11. Here the focus was to cleanup things a bit - bring Maven Wrapper and plugins up-to-date, synchronize versions and upgrade them, but keeping major versions intact, fix documentation a bit, etc.


Cali0707 (approved):
LGTM, thanks @jacekbilski !!

PR #739, opened 2026-06-09
Moving the baseline to Java 17 / Jakarta EE 10

jacekbilski:
This is the third PR of the series towards Java 17 / Spring Boot 4 / Quarkus 3 / Jakarta EE 11.

This PR is probably the biggest one of the series. We're now using Java 17 as a baseline. Also, old Jakarta EE 8 (packages starting with `javax.`) is gone, the baseline is now Jakarta EE 10. All major projects are also upgraded to their versions matching or roughly matching Jakarta EE 10. That also means that we're still using Spring Boot 3, for example.

I honestly didn't test it with Spring Boot 4 yet, but will try soon.


jacekbilski:
Hi @Cali0707, any chances you could take a look at this soon? I have already the next PR ready, see https://github.com/jacekbilski/sdk-java/pull/2.

jacekbilski:
@Cali0707, @duglin, would one of you, or someone else, be able to review this? I'd like to upgrade my app to Spring Boot 4, but this PR would be the first one to support that.

examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/JettyServer.java
```diff
@@ -33,70 +28,58 @@
...
-                            httpServletResponse.setStatus(HttpStatus.NO_CONTENT_204);
-                        }
-                    }
-                } catch (IOException e) {
-                    throw new UncheckedIOException(e);
+                if (body != null) {
+                    response.setStatus(HttpStatus.OK_200);
+                    response.write(true, ByteBuffer.wrap(body), callback);
+                } else {
+                    response.setStatus(HttpStatus.NO_CONTENT_204);
```
Cali0707, line 80: @jacekbilski I think we need to call `callback.succeeded()` here? since we don't do `response.write`

Cali0707 (commented):
Thanks for working on this @jacekbilski - this is looking really good overall

I only found one issue to look into, then let's merge this and continue along with work for the 5.0.0 release

Are there other PRs you want to land for 5.0 ?

examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/JettyServer.java
jacekbilski: Hi. Good catch. I also now asked an AI agent to review the file, and it made the same remark -> fixed.

But what it also mentioned is, that this example using Jetty 12.1 is no longer "HTTP basic" example, as it's not using standard SDK things, but Jetty's own API. Should I rename the whole module? Makes sense actually.

Cali0707: IMO the "basic http" part of this example comes from the `BasicHttpServer.java`

jacekbilski:
> Are there other PRs you want to land for 5.0 ?

As mentioned above, I have the next PR ready: https://github.com/jacekbilski/sdk-java/pull/2. There, I'm also bumping Jakarta EE version up, along with some other things like Spring Boot. I guess that would be already too much for version 5. I leave the decision to you.

examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/JettyServer.java
jacekbilski: ~Not really. In version 4.1.1 we were using for example `javax.servlet.http.HttpServletRequest` and `javax.servlet.http.HttpServletResponse`. You could call it "basic". Now it's pure Jetty.~

~See https://github.com/cloudevents/sdk-java/blob/v4.1.1/examples/basic-http/src/main/java/io/cloudevents/examples/http/basic/JettyServer.java~

~OK, I see. Now we have definitely mixed two ways together. I'll extract the Jetty module, but feel free to drop the commit before merging if it's going too far.~

It _is_ "http basic". Even if it's using some Jetty-specific API, it's also using `cloudevents-http-basic`. Separate module doesn't make sense.

Cali0707 (approved):
LGTM, thanks @jacekbilski

jacekbilski:
Hi @Cali0707, do you plan to release after having merged my PR, or are you still waiting for the next one? I'm not sure if I should already prepare it, or wait until after the release.

jacekbilski:
@Cali0707, @duglin, what are your plans right now? I thought you'd want to release this as 5.0. Then, I could create the next PR with Jakarta EE 11 / Spring Boot 4 / Jackson 3, which could become version 6.0. Or are you still waiting for something before 5.0?

I'd really like to have at least those changes officially released. Otherwise I cannot release my app with Spring Boot 4.

Cali0707:
Hey @jacekbilski sorry for the delay here - I've been slammed with other work and haven't had time to look at this recently 😞 

My current plan is to do a 5.0 release later this week, as well as the prep work required for your next PR to be part of a 6.0

Issue #704, opened 2025-12-15
CloudEvents incompatiable with Spring-framework 7.x


Stromner:
I know there is a similar issue up already about upgrading to [Spring-boot 2.7.18](https://github.com/cloudevents/sdk-java/issues/692) but the cloudevent project has lagged so far behind now that you can't even launch it with running the latest spring versions. Spring Framework 7.0 introduced a breaking change so you can no longer runt cloudevent in your project if you want to stay up to date with the rest of the ecosystem. Change in question:

> 
> **HttpHeaders changes**
> The HttpHeaders API has been revisited in 7.0. This class no longer extends the MultiValueMap contract. Underlying servers treat headers more like a collection of pairs, and many map-like operations do not behave or perform well, because headers are case-insensitive in the first place. We therefore removed several methods as a result and introduced fallbacks as immediately @Deprecated, like HttpHeaders#asMultiValueMap. Please consider other methods as much as possible. See [#33913](https://github.com/spring-projects/spring-framework/issues/33913) for more details.

Stacktrace you get when running latest release of cloudevent:
`"Class org.springframework.http.HttpHeaders does not implement the requested interface java.util.Map","error.stack_trace":"java.lang.IncompatibleClassChangeError: Class org.springframework.http.HttpHeaders does not implement the requested interface java.util.Map\n\tat io.cloudevents.http.HttpMessageFactory.lambda$createReaderFromMultimap$5(HttpMessageFactory.java:112)\n\tat io.cloudevents.http.HttpMessageFactory.createReader(HttpMessageFactory.java:70)\n\tat io.cloudevents.http.HttpMessageFactory.createReaderFromMultimap(HttpMessageFactory.java:114)\n\tat io.cloudevents.spring.http.CloudEventHttpUtils.toReader(CloudEventHttpUtils.java:55)\n\tat io.cloudevents.spring.mvc.CloudEventHttpMessageConverter.readInternal(CloudEventHttpMessageConverter.java:54)\n\tat io.cloudevents.spring.mvc.CloudEventHttpMessageConverter.readInternal(CloudEventHttpMessageConverter.java:39)\n\tat org.springframework.http.converter.AbstractHttpMessageConverter.read(AbstractHttpMessageConverter.java:196)\n\tat`

ranma2913:
https://github.com/cloudevents/sdk-java/pull/739 provided some changes.

What's left for this Issue?

I've got these issues too:

[...]

PR #743, opened 2026-07-13
🐛 Refactor `toReader` method in `CloudEventHttpUtils` to provide a temporary workaround to issue #704

ranma2913:
🐛 Refactor `toReader` method in `CloudEventHttpUtils` to provide a temporary workaround to [#704](https://github.com/cloudevents/sdk-java/issues/704)

Spring Framework 7 changed org.springframework.http.HttpHeaders so it no longer implements the Map / MultiValueMap contract.

Current versions of io.cloudevents:cloudevents-spring pass HttpHeaders directly to:

HttpMessageFactory.createReaderFromMultimap(headers, body.get()); The CloudEvents HTTP reader expects a Map<String, List<String>>, so applications running Spring Framework 7 fail while deserializing inbound CloudEvents with:

java.lang.IncompatibleClassChangeError:
Class org.springframework.http.ReadOnlyHttpHeaders does not implement the requested interface java.util.Map
In WebFlux, Reactor treats this linkage error as fatal. The HTTP exchange can remain incomplete, causing callers such as WebTestClient to report a timeout instead of the underlying exception.

### DIFF
--- a/spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
+++ b/spring/src/main/java/io/cloudevents/spring/http/CloudEventHttpUtils.java
@@ -27,8 +27,6 @@ import io.cloudevents.rw.CloudEventRWException;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.ResponseEntity;
 
-import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


Cali0707:
Hey we recently released a new version that should resolve this. Closing for now, but feel free to reopen with an explanation if this is still needed

PR #747, opened 2026-07-18
fix: handle null MessageHeaders in MessageBuilderMessageWriter (#689)

### LINKED ISSUE #689: Spring CloudEventMessageConverter produce a null pointer exception if initial message header map is null
### Step to reproduce

Declare a Rabbit templating using converter

```
@Bean
public RabbitMessagingTemplate rabbitMessagingTemplate(RabbitTemplate rabbitTemplate, CloudEventMessageConverter cloudEventMessageConverter) {
	var rmt = new RabbitMessagingTemplate(rabbitTemplate);
	rmt.setMessageConverter(cloudEventMessageConverter);
	return rmt;
}
```

Use it to publish a message
```
rabbitMessagingTemplate.convertAndSend("my.routing.key", CloudEventBuilder.v1()
				.withId(UUID.randomUUID().toString())
				.withType("mytype")
				.withSource(new URI("my:urisource"))
				.withData(new BytesCloudEventData("{\"data\":\"test\"}".getBytes()))
				.build());
```

### Result
NPE
```
java.lang.NullPointerException: Cannot invoke "java.util.Map.size()" because "m" is null

	at java.base/java.util.HashMap.putMapEntries(HashMap.java:495)
	at java.base/java.util.HashMap.putAll(HashMap.java:783)
	at io.cloudevents.spring.messaging.MessageBuilderMessageWriter.<init>(MessageBuilderMessageWriter.java:45)
	at io.cloudevents.spring.messaging.CloudEventMessageConverter.toMessage(CloudEventMessageConverter.java:54)
	at org.springframework.messaging.core.AbstractMessageSendingTemplate.doConvert(AbstractMessageSendingTemplate.java:177)
	at org.springframework.messaging.core.AbstractMessageSendingTemplate.convertAndSend(AbstractMessageSendingTemplate.java:150)
	at org.springframework.messaging.core.AbstractMessageSendingTemplate.convertAndSend(AbstractMessageSendingTemplate.java:129)
	at org.springframework.messaging.core.AbstractMessageSendingTemplate.convertAndSend(AbstractMessageSendingTemplate.java:122)
```

### Solution

In call to `io.cloudevents.spring.messaging.CloudEventMessageConverter#toMessage`, the provided headers field can be null if caller didn't provide custom headers. A null check should be done and MessabeBilderMessageWriter should be called with default constructor in that case.

### Workaround

use RabbitMessagingTemplate send methods that take a map as parameter

prashantdotnaagar:
## Summary

Fixes a `NullPointerException` that occurs when `CloudEventMessageConverter.toMessage()` is called with `null` `MessageHeaders` — a scenario that happens when using `RabbitMessagingTemplate#convertAndSend(String, Object)` (and similar Spring messaging templates that don't supply headers).

## Root Cause

`CloudEventMessageConverter.toMessage()` passes the `headers` argument directly to `new MessageBuilderMessageWriter(headers)`. Inside the constructor, `this.headers.putAll(headers)` was called unconditionally, throwing an NPE when `headers` is `null`.

## Fix

Added a null-guard in the `MessageBuilderMessageWriter(Map<String, Object> headers)` constructor so that a `null` input is treated as an empty map, while CloudEvent attributes are still written correctly.

## Test

Added regression test `fromCloudEventWithNullHeaders` to `CloudEventMessageConverterTests` that calls `converter.toMessage(event, null)` and asserts the returned message contains the correct CloudEvent headers.

All 9 tests in `CloudEventMessageConverterTests` pass.

Closes #689

### DIFF
--- a/spring/src/test/java/io/cloudevents/spring/messaging/CloudEventMessageConverterTests.java
+++ b/spring/src/test/java/io/cloudevents/spring/messaging/CloudEventMessageConverterTests.java
@@ -128,4 +128,19 @@ class CloudEventMessageConverterTests {
 		assertThat(converter.toMessage(new byte[0], new MessageHeaders(Collections.emptyMap()))).isNull();
 	}
 
+	// gh-689: toMessage should not throw NPE when the caller provides no headers,
+	// e.g. RabbitMessagingTemplate#convertAndSend(String, Object) passes null.
+	@Test
+	void fromCloudEventWithNullHeaders() {
+		CloudEvent attributes = CloudEventBuilder.v1().withId("A234-1234-1234")
+				.withSource(URI.create("https://spring.io/")).withType("org.springframework")
+				.withData("hello".getBytes(StandardCharsets.UTF_8)).build();
+		Message<?> message = converter.toMessage(attributes, null);
+		Map<String, ?> headers = message.getHeaders();
+		assertThat(headers.get("ce-id")).isEqualTo("A234-1234-1234");
+		assertThat(headers.get("ce-specversion")).isEqualTo("1.0");
+		assertThat(headers.get("ce-source")).isEqualTo("https://spring.io/");
+		assertThat(headers.get("ce-type")).isEqualTo("org.springframework");
+	}
+
 }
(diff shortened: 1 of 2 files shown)


spring/src/test/java/io/cloudevents/spring/messaging/CloudEventMessageConverterTests.java
```diff
@@ -128,4 +128,19 @@ void fromNonCloudEvent() {
 		assertThat(converter.toMessage(new byte[0], new MessageHeaders(Collections.emptyMap()))).isNull();
 	}
 
+	// gh-689: toMessage should not throw NPE when the caller provides no headers,
+	// e.g. RabbitMessagingTemplate#convertAndSend(String, Object) passes null.
```
Cali0707, line 132: Nit: let's either:
1. not ref the issue in the test case
2. actually link to the issue rather than using a shorthand

+1 on having the test case though

